PR #7, opened 2018-11-26
early start and enable hostNetwork

sjenning:
50 is machine-api-operator, which we will need to do CSR validation, so I choose 51.

`hostNetwork` is so the pod can schedule on nodes that are not ready due to CNI being uninitialized.

@smarterclayton can I use my in-cluster kubeconfig if I'm running hostNetwork i.e. is the kubernetes svc endpoint resolvable?  

fyi @mrogers950


smarterclayton:
No, you have to do what the other control plane elements do (set KUBERNETES_SERVICE_PORT/HOST to loopback).

PR #8, opened 2018-12-19
Operator restructure

mrogers950:
Use the cluster-osin-operator repo layout and boilerplate (with minor modifications).
@openshift/sig-auth

PR #9, opened 2018-12-22
Revert "Operator restructure"

smarterclayton:
Reverts openshift/cluster-machine-approver#8

Post merge seeing a large number of e2e-aws failures across all repos on machine-approver-controller restarting.


smarterclayton:
@wking if this greens I'm going to revert

abhinavdahiya:
@smarterclayton this is green

wking:
I dropped some log excerpts [here][1].

[1]: https://github.com/openshift/installer/pull/955#issuecomment-449532589

enj:
Hmm restart technically occurrs because it lost leader election.  We could turn that off I guess.  The 127.0.0.1 ip also looks weird.

smarterclayton:
Logs are in all the linked jobs

PR #10, opened 2019-01-02
Approve CSRs based on machine status

mrogers950:


main.go
```diff
@@ -99,7 +103,29 @@ func (c *Controller) handleNewCSR(key string) error {
 		return nil
 	}
 
-	// TODO and CSR checking logic here
+	var machineListObj runtime.Object
+	err = c.clientset.RESTClient().Get().Namespace("openshift-cluster-api").Name("machines").Do().Into(machineListObj)
```
sjenning: should we approve if the machine API isn't available?

sjenning: we can check the username `system:node[s]...`  other than that, i don't think there is much we can check.

main.go
```diff
@@ -99,12 +103,43 @@ func (c *Controller) handleNewCSR(key string) error {
...
-	// TODO and CSR checking logic here
+	parsedCSR, err := csrclient.ParseCSR(csr)
+	if err != nil {
+		glog.Infof("error parsing request CSR: %v", err)
+		return nil
+	}
+
+	approvalMsg := "This CSR was approved by the Node CSR Approver"
+	var machineListObj runtime.Object
+	err = c.clientset.RESTClient().Get().Namespace("openshift-cluster-api").Name("machines").Do().Into(machineListObj)
```
mrogers950: I just need to add in the actual client for this part.

derekwaynecarr:
fyi @enxebre

PR #12, opened 2019-01-21
Change run level to 09

enj:
Signed-off-by: Monis Khan <mkhan@redhat.com>

/assign @smarterclayton @deads2k 

@openshift/sig-auth 

Based on discussion in slack to help with CI wedging.


smarterclayton:
will tag when it goes green

PR #14, opened 2019-03-06
Add CSR failure cause to logs

cgwalters:
My worker's CSRs aren't being approved and this would help me
debug why.

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto/x509"
+	"fmt"
 	"strings"
 
 	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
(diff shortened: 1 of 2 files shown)


cgwalters:
(I didn't test this beyond running `make` though)

cgwalters:
Deployed this in my local libvirt cluster, and now I see:
```
I0307 15:17:49.020369       1 main.go:123] CSR csr-fvcvt not authorized: Missing SAN in DNSNames
```

csr_check.go
```diff
@@ -127,7 +128,7 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
 		}
 		// The CSR requested a DNS name that did not belong to the machine
 		if !foundSan {
-			return false
+			return fmt.Errorf("Missing SAN in DNSNames")
```
mrogers950: This error message is not entirely accurate, the error case here is where the CSR has requested a SAN that is not present in the machine's addresses. It might be helpful to print out the CSR and machine SANs also.

csr_check.go
```diff
@@ -147,9 +148,9 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
 		}
 		// The CSR requested an IP name that did not belong to the machine
 		if !foundSan {
-			return false
+			return fmt.Errorf("Missing SAN in IPAddresses")
```
mrogers950: Same as above comment.

csr_check.go
cgwalters: Done :arrow_down:

cgwalters:
New error is:

```
I0307 16:03:37.622478       1 main.go:97] CSR csr-w4pq8 added
I0307 16:03:37.629018       1 main.go:123] CSR csr-w4pq8 not authorized: DNS name 'osiris-q4f59-worker-0-rj26d' not in machine names: 
```

Which, looking at `oc -n openshift-machine-api edit machines/osiris-q4f59-worker-0-rj26d` indeed I just see:

```
status:
  addresses:
  - address: 192.168.126.51
    type: InternalIP
```

cgwalters:
And on AWS:

```
status:
  addresses:
  - address: 10.0.139.202
    type: InternalIP
  - address: ""
    type: ExternalDNS
  - address: ip-10-0-139-202.ec2.internal
    type: InternalDNS
```

PR #18, opened 2019-04-22
Bump deps and fix unit tests

sallyom:
@mrogers950 @enj  
vendored pkgs match library-go's


csr_check_test.go
```diff
@@ -1024,16 +1023,19 @@ func TestAuthorizeCSR(t *testing.T) {
...
 		if err != nil {
 			t.Fatalf("%s: error parsing test input csr %v", name, err)
 		}
-		if result := authorizeCSR(tc.machineList, tc.request, parsedCSR); result != tc.expected {
-			t.Fatalf("%s: expected %v, got %v", name, tc.expected, result)
+		err = authorizeCSR(tc.machineList, tc.request, parsedCSR)
+		if tc.expected {
+			if err != nil {
+				t.Fatalf("%v: error authorizing CSR", err)
+			}
```
sallyom: probably a better way to fix these, but i made a quick fix here to get e2e-aws rolling

sallyom:
not the flake i was looking for...
/retest

csr_check_test.go
mrogers950: This doesn't look right, our check should be
```
expected == false
     if err == nil, fatal
else
expected == true
     if err != nil, fatal
```

mrogers950:
Looks good, I just have the one comment on the test.

sallyom:
@mrogers950 ptal again, thank you!

enj:
Release repo needs to be updated to actually run the unit tests.

glide.yaml
```diff
@@ -1,25 +1,24 @@
...
+  version: kubernetes-1.14.0
   subpackages:
   - kubernetes
   - tools/cache
   - tools/clientcmd
   - util/workqueue
-- package: github.com/golang/glog
 - package: github.com/openshift/cluster-api
-  version: release-4.0
+  version: openshift-4.2-cluster-api-0.0.0-alpha.4
```
sallyom: @enj @mrogers950 FWIU this branch has updates/fixes we want/need?

sallyom:
> Release repo needs to be updated to actually run the unit tests.
https://github.com/openshift/release/pull/3841

glide.yaml
enj: We want to stay in sync with the branch that lines up with the release of OpenShift we are shipping.  I am OK if we say `master` is good enough for the cluster API code (it probably is).  The kube code is the special unicorn that we need to pin to a specific release branch.

csr_check_test.go
```diff
@@ -1963,7 +1962,11 @@ func Test_authorizeCSR(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			tt.args.req.Spec.Request = []byte(tt.args.csr)
-			parsedCSR, err := csr.ParseCSR(tt.args.req)
+			parsedCSR, err := parseCSR(&certificatesv1beta1.CertificateSigningRequest{
```
enj: This should just be a one line change.

glide.yaml
```diff
@@ -1,25 +1,24 @@
...
 - package: sigs.k8s.io/controller-runtime
```
enj: Still need to drop.

sallyom: _oh_ last comment I _thought_  you meant me to drop the cluster-api above, then was confused bc we're calling that, so then thought you meant to drop my _change_ to the version.  Now I see you meant to drop the controller-runtime, bc we are no longer calling that anywhere! got it!

glide.yaml
sallyom: I've changed to master

csr_check_test.go
sallyom: yes, fixed

PR #21, opened 2019-04-26
return the error to allow for machineset retries

rphillips:
The MachineSet List is only attempted once due to a nil being returned to forget the key on an error path. This patch allows the controller to attempt to authorize the CSR (5 times).

Possibly Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1687295#c9

@sallyom @mrogers950

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,7 +25,6 @@ import (
 
 	//"sigs.k8s.io/controller-runtime/pkg/client/config"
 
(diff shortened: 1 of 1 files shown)


rphillips:
Dupe https://github.com/openshift/cluster-machine-approver/pull/22

PR #26, opened 2019-05-10
Handle node client certs

enj:
Signed-off-by: Monis Khan <mkhan@redhat.com>

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -4,10 +4,15 @@ import (
 	"crypto/x509"
 	"fmt"
 	"strings"
+	"time"
 
 	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 )
@@ -16,27 +21,41 @@ const (
 	nodeUser       = "system:node"
 	nodeGroup      = "system:nodes"
 	nodeUserPrefix = nodeUser + ":"
+
+	maxPendingDelta = time.Hour
+	maxPendingCSRs  = 100
+
+	nodeBootstrapperUsername = "system:serviceaccount:openshift-machine-config-operator:node-bootstrapper"
+
+	maxMachineClockSkew = 10 * time.Second
+	maxMachineDelta     = 10 * time.Minute
+)
+
+var nodeBootstrapperGroups = sets.NewString(
+	"system:serviceaccounts:openshift-machine-config-operator",
+	"system:serviceaccounts",
+	"system:authenticated",
 )
 
 func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr *x509.CertificateRequest) (string, error) {
 	if !strings.HasPrefix(req.Spec.Username, nodeUserPrefix) {
-		return "",fmt.Errorf("Doesn't match expected prefix")
+		return "", fmt.Errorf("Doesn't match expected prefix")
 	}
 
 	nodeAsking := strings.TrimPrefix(req.Spec.Username, nodeUserPrefix)
-	if len(nodeAsking) < 1 {
-		return "",fmt.Errorf("Empty name")
+	if len(nodeAsking) == 0 {
+		return "", fmt.Errorf("Empty name")
 	}
 
 	// Check groups, we need at least:
 	// - system:nodes
 	// - system:authenticated
 	if len(req.Spec.Groups) < 2 {
-		return "",fmt.Errorf("Too few groups")
+		return "", fmt.Errorf("Too few groups")
 	}
 	groupSet := sets.NewString(req.Spec.Groups...)
 	if !groupSet.HasAll(nodeGroup, "system:authenticated") {
-		return "",fmt.Errorf("Not in system:authenticated")
+		return "", fmt.Errorf("Not in system:authenticated")
 	}
 
 	// Check usages, we need only:
@@ -44,7 +63,7 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
 	// - key encipherment
 	// - server auth
 	if len(req.Spec.Usages) != 3 {
-		return "",fmt.Errorf("Too few usages")
+		return "", fmt.Errorf("Too few usages")
 	}
 
 	usages := make([]string, 0)
@@ -54,7 +73,7 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
 
 	// No extra usages!
 	if len(usages) != 3 {
-		return "",fmt.Errorf("Unexpected usages: %d", len(usages))
+		return "", fmt.Errorf("Unexpected usages: %d", len(usages))
 	}
 
 	usageSet := sets.NewString(usages...)
@@ -63,12 +82,12 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
 		string(certificatesv1beta1.UsageKeyEncipherment),
 		string(certificatesv1beta1.UsageServerAuth),
 	) {
-		return "",fmt.Errorf("Missing usages")
+		return "", fmt.Errorf("Missing usages")
 	}
 
 	// Check subject: O = system:nodes, CN = system:node:ip-10-0-152-205.ec2.internal
 	if csr.Subject.CommonName != req.Spec.Username {
-		return "",fmt.Errorf("Mismatched CommonName %s != %s", csr.Subject.CommonName, req.Spec.Username)
+		return "", fmt.Errorf("Mismatched CommonName %s != %s", csr.Subject.CommonName, req.Spec.Username)
 	}
 
 	var hasOrg bool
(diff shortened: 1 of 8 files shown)


node_client.go
```diff
@@ -0,0 +1,52 @@
...
+	capi.UsageKeyEncipherment,
+	capi.UsageDigitalSignature,
+	capi.UsageClientAuth,
+}
+
+func isNodeClientCert(csr *capi.CertificateSigningRequest, x509cr *x509.CertificateRequest) bool {
+	if !reflect.DeepEqual([]string{"system:nodes"}, x509cr.Subject.Organization) {
+		return false
+	}
+	if (len(x509cr.DNSNames) > 0) || (len(x509cr.EmailAddresses) > 0) || (len(x509cr.IPAddresses) > 0) {
```
mrogers950: If the cert has the exact usages then we don't need to do this check.

main.go
```diff
@@ -114,31 +112,21 @@ func (c *Controller) handleNewCSR(key string) error {
...
 	if err != nil {
-		glog.Infof("machine api not available: %v", err)
-		// Validate the CSR for the bootstrapping phase without a SAN check.
-		_, err := validateCSRContents(csr, parsedCSR)
-		if err != nil {
-			glog.Infof("CSR %s not valid: %v", csr.GetName(), err)
-			return err
-		}
-		approvalMsg += " (no SAN validation)"
+		return fmt.Errorf("failed to list machines: %v", err)
```
mrogers950: Do we want to do this now? testing this PR would then be contingent on both the bootstrap self-approve script and PR to disable the auto-approve. just wondering

csr_check.go
```diff
@@ -162,3 +180,74 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
...
+	}
+
+	nodeMachine, ok := findMatchingMachineFromInternalDNS(nodeName, machines)
+	if !ok {
+		return fmt.Errorf("failed to find machine for node %s", nodeName)
+	}
+
+	if nodeMachine.Status.NodeRef != nil {
+		return fmt.Errorf("machine for node %s already has node ref", nodeName)
+	}
```
mrogers950: Does finding the nodeMachine and checking for an empty ref satisfy: 5) Ensure there’s machine object that triggered this new node trying to join the cluster. ?

node_client.go
enj: Whole file is copied from the sar approver.  I just left it alone.

main.go
enj: If it passes CI we know that we don't need the failover, right?

csr_check.go
enj: I would say `findMatchingMachineFromInternalDNS` does that by itself.

enj: (granted you could say all of the machine checks are required to fulfill the requirement)

main.go
enj: Also, you could get a transient error and hit the failover path because the code is too blunt.  I think we need to remove it before GA so might as well do it now.

mrogers950: > If it passes CI we know that we don't need the failover, right?

Yeah, otherwise removing it might stop us from bootstrapping with no machine API available. So let's see what CI says.. I am +1 on getting rid of the failover here though.

node_client.go
mrogers950: ok

enj:
Once we got past the AWS issues this ran fine.  Now I just need to add more unit tests.

enj:
Well at least the serving certs bit is working fine.  The client cert is failing on RBAC:

```
I0511 12:39:05.629859       1 main.go:122] CSR csr-5696g not authorized: failed to check if node ip-10-0-146-108.ec2.internal already exists: nodes "ip-10-0-146-108.ec2.internal" is forbidden: User "system:serviceaccount:openshift-cluster-machine-approver:machine-approver-sa" cannot get resource "nodes" in API group "" at the cluster scope
```

csr_check.go
```diff
@@ -162,3 +180,77 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
...
+	for _, machine := range machines {
+		if machine.Status.NodeRef != nil && machine.Status.NodeRef.Name == nodeName {
+			return machine, true
+		}
+	}
+	return v1beta1.Machine{}, false
+}
+
+func findMatchingMachineFromInternalDNS(nodeName string, machines []v1beta1.Machine) (v1beta1.Machine, bool) {
+	for _, machine := range machines {
```
enxebre: I'm not sure this is something we can guarantee being available for every cloud. Not sure `NodeInternalDNS` is exposed on the API in Azure/GCP. It might be only exposed via metadata.

csr_check.go
```diff
@@ -162,3 +180,77 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
...
+	nodeMachine, ok := findMatchingMachineFromInternalDNS(nodeName, machines)
+	if !ok {
+		return fmt.Errorf("failed to find machine for node %s", nodeName)
+	}
+
+	if nodeMachine.Status.NodeRef != nil {
+		return fmt.Errorf("machine for node %s already has node ref", nodeName)
+	}
+
+	if !isWorkerMachine(nodeMachine) {
```
enxebre: I understand we don't want masters to go through this however is this effectively dropping machines which does not contain https://github.com/openshift/cluster-machine-approver/pull/26/files#diff-a482a1a664767249971b0b71ca9baaf0R28? How about any additional user machines/machineSets? I wonder if we might want to have an specific label for this e.g `machine.openshift.io/CSR: true` or possible better the other way around and have masters to opt out with `machine.openshift.io/CSR-need-approval: false`

csr_check.go
enj: Any other way to very specifically find the machine?

enxebre: would the privateIP be sufficient?
medium term we should include the node/machine providerID here but this is not supported today

csr_check.go
enj: I wanted the most strict whitelist based approach.  This approval flow is only meant for workers so this check seemed appropriate.  I am not for or against adding a new label, we just need to discuss that type of change with the other stakeholders.

csr_check.go
enj: The CSR only includes the (future) node name so we have to base the check on that...

enxebre: couldn't we get the privateIP here (in future providerID as well) https://github.com/openshift/cluster-machine-approver/pull/26/files#diff-a482a1a664767249971b0b71ca9baaf0R194 and pass it down to `findMatchingMachineBy*()`?

csr_check.go
```diff
@@ -88,39 +106,39 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
 // authorizeCSR authorizes the CertificateSigningRequest req for a node's server certificate.
```
mrogers950: This comment needs to be updated

csr_check_test.go
```diff
@@ -1004,36 +1156,906 @@ func TestAuthorizeCSR(t *testing.T) {
...
+				},
+				nodeName: "panda",
+				nodeErr:  errors.NewNotFound(schema.GroupResource{}, ""),
+				req: &certificatesv1beta1.CertificateSigningRequest{
+					Spec: certificatesv1beta1.CertificateSigningRequestSpec{
+						Usages: []certificatesv1beta1.KeyUsage{
+							certificatesv1beta1.UsageKeyEncipherment,
+							certificatesv1beta1.UsageDigitalSignature,
+							certificatesv1beta1.UsageClientAuth,
+							certificatesv1beta1.UsageServerAuth,
```
mrogers950: We should have variants that have less usages and wrong ones (that replace client with server) for good measure.

csr_check.go
```diff
@@ -162,3 +180,74 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
...
+
+	if nodeMachine.Status.NodeRef != nil {
+		return fmt.Errorf("machine for node %s already has node ref", nodeName)
+	}
+
+	if !isWorkerMachine(nodeMachine) {
+		return fmt.Errorf("machine for node %s is not a worker", nodeName)
+	}
+
+	start := nodeMachine.CreationTimestamp.Add(-maxMachineDelta) // TODO maybe this should be really small to account only for clock drift
```
mrogers950: Reduce as discussed on call: 5-10 sec

csr_check.go
```diff
@@ -162,3 +180,74 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
...
+			if address.Type == corev1.NodeInternalDNS && address.Address == nodeName {
+				return machine, true
+			}
+		}
+	}
+	return v1beta1.Machine{}, false
+}
+
+func isWorkerMachine(machine v1beta1.Machine) bool {
+	return machine.Labels[machineRoleKey] == machineRoleTypeValue && machine.Labels[machineTypeKey] == machineRoleTypeValue
```
mrogers950: Change to the to-be-agreed-upon label

mrogers950:
@enj looks good, squash the commits and I'll tag.

csr_check.go
```diff
@@ -79,48 +101,61 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
...
+// csr should be the parsed CSR from req.Spec.Request.
+//
+// For client certificates:
+// The only information contained in the CSR is the future name of the node.  Thus we perform a best effort check:
+//
+// 1. User is the node bootstrapper
+// 2. Node does not exist
+// 3. Use machine API internal DNS to locate matching machine based on node name
+// 4. Machine must not have a node ref
+// 5. Machine must opt-in to CSR auto approve via label
```
derekwaynecarr: can we weigh pros/cons of label versus annotation?

csr_check.go
```diff
@@ -162,3 +197,110 @@ func authorizeCSR(machineList *v1beta1.MachineList, req *certificatesv1beta1.Cer
...
+			if address.Type == corev1.NodeInternalDNS && address.Address == nodeName {
+				return machine, true
+			}
+		}
+	}
+	return v1beta1.Machine{}, false
+}
+
+func machineWantsCSRAutoApprove(machine v1beta1.Machine) bool {
+	return machine.Labels[machineCSRAutoApproveKey] == machineCSRAutoApproveValue
```
derekwaynecarr: just want to know pros/cons of label versus annotation.

i tend to think of labels as grouping, and annotations as behavioral.

enj:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: machine-approver-config
  namespace: openshift-cluster-machine-approver
data:
  config.yaml: |-
    nodeClientCert:
      disabled: true
```

```
I0515 18:57:00.248954       1 config.go:23] machine approver config: {NodeClientCert:{Disabled:true}}
I0515 18:57:00.249006       1 main.go:183] Starting Machine Approver
I0515 18:57:25.336449       1 main.go:107] CSR csr-cwcjt added
I0515 18:57:25.344832       1 main.go:132] CSR csr-cwcjt not authorized: CSR csr-cwcjt for node client cert rejected as the flow is disabled
I0515 18:57:25.344856       1 main.go:164] Error syncing csr csr-cwcjt: CSR csr-cwcjt for node client cert rejected as the flow is disabled
```

csr_check.go
```diff
@@ -16,35 +21,49 @@ const (
...
 	nodeGroup      = "system:nodes"
 	nodeUserPrefix = nodeUser + ":"
+
+	maxPendingDelta = time.Hour
+	maxPendingCSRs  = 100
+
+	nodeBootstrapperUsername = "system:serviceaccount:openshift-machine-config-operator:node-bootstrapper"
+
+	maxMachineClockSkew = 10 * time.Second
+	maxMachineDelta     = 10 * time.Minute
```
zeenix: May I ask where this choice of number comes from? In our CI system (nested virt env), we seem to be quite a bit late:

```
E0701 18:15:26.348374       1 main.go:174] CSR csr-xl8fs creation time 2019-07-01 18:15:26 +0000 UTC not in range (2019-07-01 17:40:40 +0000 UTC, 2019-07-01 17:50:50 +0000 UTC)
```

enj: We are not performing any cryptographic verification.  Instead we are making a best effort check based on the state of the machine API.  Thus we need to keep the window very short to prevent the chance of an attacker guessing the correct name of a new node.  Increasing the window gives the attacker more time and lowers the security of this approver.

zeenix: @enj Thanks for explaining. However the time of choice, seems very arbitrary to me. Please correct me if I'm wrong since I'm no security export but isn't 10mis a very large window already for an attacker? Does this check really add any real security to the system?

enj: We also rate limit based on the total amount of outstanding CSRs.  I would not call it great security, but we make it hard enough to exploit for the time being.

Note that the solution is not to increase the window.  The correct solution is to cryptographically tie the CSR with the new node that is being provisioned.  Then an attacker cannot create a CSR to steal credentials.  It would also allow nodes to create more than one CSR (it allows them to not store state about previously requested CSRs).

zeenix: Thanks again for your quick reply and detailed explanation.

I'm actually more suggesting if the window is needed at all. Is the 10 mins time short enough for attackers to not be able to exploit? If it's not, I would rather suggest this check be removed as it's causing us problems and not helping in that case. The real solution you mention, sounds good and it won't affect us negatively.

enj: You have to know when that 10 min windows occurs - it is actually pretty hard to know without access to the machine API or the underlying cloud API.  That basically leaves brute force methods, which the rate limiting does a good job of preventing.  Without the window you could just lay down a handful of guess CSRs and see if you get lucky.  The windows forces a much more involved attack.  Basically it forces an active attacker instead of a passive one.  The implementations of the machine API need to properly coordinate the creation of the machine object and the related CSR object.

Feel free to bring this up in the ABC arch call for further discussion if desired.

zeenix: Thanks again for another very thorough and detailed answer. Much appreciated!

You'll have to understand that this breaks our (Openshift Installer on libvirt) CI right now. It's a nested environment so understandably things are slower than normal, despite the fact that the VM has 8 core, 32G RAM and and SSD drive assigned to it. Apart from arguing for removal or extension of the time window check, if you have any other suggestions on how to fix or workaround this issue, please do let me know cause this is very high priority for us (multiple teams).

I'm afraid I don't know what ABC is. I'm still relatively new here but I'd love to join if we could discuss it there.  :)

PR #31, opened 2019-07-10
Max CSR limit

enxebre:
Currently the machine approver leave CSR pending if there's not legit machine for the node bootstrap CSR.
if pending CSR >100 the auto approver decides refuses to approve any CSR thereafter.
This can easily happen in a legit rapid scale up since each kubelet will create 2 CSRs
https://bugzilla.redhat.com/show_bug.cgi?id=1717610

This PR would deny CSRs for node bootstrapping with no legit backing machine and use denied CSRs as the threshold for blocking the approver

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -3,6 +3,7 @@ package main
 import (
 	"crypto/x509"
 	"fmt"
+	"k8s.io/klog"
 	"strings"
 	"time"
 
@@ -22,8 +23,8 @@ const (
 	nodeGroup      = "system:nodes"
 	nodeUserPrefix = nodeUser + ":"
 
-	maxPendingDelta = time.Hour
-	maxPendingCSRs  = 100
+	maxDeniedDelta = time.Hour
+	maxDeniedCSRs  = 100
 
 	nodeBootstrapperUsername = "system:serviceaccount:openshift-machine-config-operator:node-bootstrapper"
 
@@ -273,13 +274,22 @@ func isApproved(csr *certificatesv1beta1.CertificateSigningRequest) bool {
 	return false
 }
 
-func recentlyPendingCSRs(indexer cache.Indexer) int {
+func isDenied(csr *certificatesv1beta1.CertificateSigningRequest) bool {
+	for _, condition := range csr.Status.Conditions {
+		if condition.Type == certificatesv1beta1.CertificateDenied {
+			return true
+		}
+	}
+	return false
+}
+
+func recentlyDeniedCSRs(indexer cache.Indexer) int {
 	// assumes we are scheduled on the master meaning our clock is the same
 	now := time.Now()
-	start := now.Add(-maxPendingDelta)
+	start := now.Add(-maxDeniedDelta)
 	end := now.Add(maxMachineClockSkew)
 
-	var pending int
+	var denied int
 
 	for _, item := range indexer.List() {
 		csr := item.(*certificatesv1beta1.CertificateSigningRequest)
@@ -289,10 +299,20 @@ func recentlyPendingCSRs(indexer cache.Indexer) int {
 			continue
 		}
 
-		if !isApproved(csr) {
-			pending++
+		if isDenied(csr) {
+			denied++
 		}
 	}
 
-	return pending
+	return denied
+}
+
+func isNodeBootstrapCSR(csr *certificatesv1beta1.CertificateSigningRequest) bool {
+	parsedCSR, err := parseCSR(csr)
+	if err != nil {
+		klog.Errorf("Failed to parse CSR: %v", err)
+		return false
+	}
+	nodeAsking, _:= validateCSRContents(csr, parsedCSR)
+	return isNodeClientCert(csr, parsedCSR) || nodeAsking != ""
 }
--- a/main.go
+++ b/main.go
@@ -209,6 +209,7 @@ func main() {
 		cliConfig  string
 	)
 
+	klog.InitFlags(nil)
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "absolute path to the kubeconfig file")
 	flag.StringVar(&master, "master", "", "master url")
 	flag.StringVar(&cliConfig, "config", "", "CLI config")
--- a/main.go
+++ b/main.go
@@ -106,19 +106,24 @@ func (c *Controller) handleNewCSR(key string) error {
 	// is dependent on the actual instance, to detect that a CSR was recreated with the same name
 	klog.Infof("CSR %s added", csr.Name)
 
+	parsedCSR, err := parseCSR(csr)
+	if err != nil {
+		klog.Infof("error parsing request CSR: %v", err)
+		return nil
+	}
+
+	if !isNodeBootstrapCSR(csr) {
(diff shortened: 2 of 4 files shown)


enxebre:
/hold
cc @sjenning @sttts @enj @mrunalp

csr_check.go
```diff
@@ -289,10 +299,20 @@ func recentlyPendingCSRs(indexer cache.Indexer) int {
...
+		if isDenied(csr) {
+			denied++
 		}
 	}
 
-	return pending
+	return denied
+}
+
+func isNodeBootstrapCSR(csr *certificatesv1beta1.CertificateSigningRequest) bool {
```
frobware: This function should return an error. `(bool, error)`. Let the caller decide what to do if there is an error.

csr_check.go
```diff
@@ -289,10 +299,20 @@ func recentlyPendingCSRs(indexer cache.Indexer) int {
...
+	parsedCSR, err := parseCSR(csr)
+	if err != nil {
+		klog.Errorf("Failed to parse CSR: %v", err)
+		return false
+	}
+	nodeAsking, _:= validateCSRContents(csr, parsedCSR)
```
frobware: Why do we ignore any potential error here?

mrogers950:
Why use denials instead of an annotation on the CSR for the threshold? If the approver starts denying, it would be the sole ruler of the CSR API, and there would not be room for a future secondary approval controller to approve a different kind of CSR. We might not ever need that though, just pointing it out.

main.go
```diff
@@ -106,19 +106,24 @@ func (c *Controller) handleNewCSR(key string) error {
...
-		klog.Infof("CSR %s is already approved", csr.Name)
+	parsedCSR, err := parseCSR(csr)
+	if err != nil {
+		klog.Infof("error parsing request CSR: %v", err)
 		return nil
 	}
 
-	if pending := recentlyPendingCSRs(c.indexer); pending > maxPendingCSRs {
-		klog.Infof("ignoring all CSRs as too many recent pending CSRs seen: %d", pending)
+	if !isNodeBootstrapCSR(csr) {
```
mrunalp: Aren't we skipping the node serving cert CSRs entirely with this check moved up here? 
(The code path from down here - https://github.com/openshift/cluster-machine-approver/blob/master/csr_check.go#L131)

csr_check.go
vikaschoudhary16: we should log warning i think. It will help in debugging if somehow approver is skipping the csrs which it should not be.

main.go
vikaschoudhary16: +1, may be we should refactor `authorizeCSR()` and invoke `isNodeServingCert()` from here and from within authorizeCSR()

main.go
```diff
@@ -106,19 +106,24 @@ func (c *Controller) handleNewCSR(key string) error {
...
-	parsedCSR, err := parseCSR(csr)
-	if err != nil {
-		klog.Infof("error parsing request CSR: %v", err)
+	if isApproved(csr) {
+		klog.Infof("CSR %s is already approved", csr.Name)
+		return nil
+	}
+
+	if denied := recentlyDeniedCSRs(c.indexer); denied > maxDeniedCSRs {
+		klog.Infof("ignoring all CSRs as too many recent denied CSRs seen: %d", denied)
```
vikaschoudhary16: s/ignoring/Ignoring

vikaschoudhary16:
Instead of denials, how about increasing pending csr counter only for node client and serving certs?

csr_check.go
```diff
@@ -22,8 +23,8 @@ const (
 	nodeGroup      = "system:nodes"
 	nodeUserPrefix = nodeUser + ":"
 
-	maxPendingDelta = time.Hour
-	maxPendingCSRs  = 100
+	maxDeniedDelta = time.Hour
```
vikaschoudhary16: How about increasing `maxPendingDelta` only for the the node client and serving certs? This way we wont deny any CSRs unwantedly.

enxebre:
> Why use denials instead of an annotation on the CSR for the threshold? If the approver starts denying, it would be the sole ruler of the CSR API, and there would not be room for a future secondary approval controller to approve a different kind of CSR. We might not ever need that though, just pointing it out.

@mrogers950 this is attempt to mitigate what's stated in the description https://bugzilla.redhat.com/show_bug.cgi?id=1717610

main.go
enxebre: > Aren't we skipping the node serving cert CSRs entirely with this check moved up here?
> (The code path from down here - https://github.com/openshift/cluster-machine-approver/blob/master/csr_check.go#L131)

this checks if this is a node bootstrappping CSR for either client or server. If it is not, skip so others approvers can react.

vikaschoudhary16: For server, 
> Names contained in the CSR are checked against addresses in the corresponding node's machine status
```golang
 // Check that we have a registered node with the request name
        targetMachine, ok := findMatchingMachineFromNodeRef(nodeAsking, machines)
        if !ok {
                return fmt.Errorf("No target machine")
        }

        // SAN checks for both DNS and IPs, e.g.,
        // DNS:ip-10-0-152-205, DNS:ip-10-0-152-205.ec2.internal, IP Address:10.0.152.205, IP Address:10.0.152.205
        // All names in the request must correspond to addresses assigned to a single machine.
        for _, san := range csr.DNSNames {
                if len(san) == 0 {
                        continue
                }
                var attemptedAddresses []string
                var foundSan bool
                        switch addr.Type {
                        case corev1.NodeInternalDNS, corev1.NodeExternalDNS, corev1.NodeHostName:
                                if san == addr.Address {
                                        foundSan = true
                                        break
                                } else {
                                        attemptedAddresses = append(attemptedAddresses, addr.Address)
                                }
                        default:
                        }
                }
                if !foundSan {
                        return fmt.Errorf("DNS name '%s' not in machine names: %s", san, strings.Join(attemptedAddresses, " "))
                }
        }

        for _, san := range csr.IPAddresses {
                if len(san) == 0 {
                        continue
                }
                var attemptedAddresses []string
                var foundSan bool
                        switch addr.Type {
                        case corev1.NodeInternalIP, corev1.NodeExternalIP:
                                if san.String() == addr.Address {
                                        foundSan = true
                                        break
                                } else {
                                        attemptedAddresses = append(attemptedAddresses, addr.Address)
                                }
                        default:
                        }
                }
                if !foundSan {
                        return fmt.Errorf("IP address '%s' not in machine addresses: %s", san, strings.Join(attemptedAddresses, " "))
                }
        }

```

enj:
I disagree with Seth's suggestion for a deny based approach.  Per Matt's comment, it makes it harder to write multiple approvers.

I propose a simpler approach - to prevent dictionary style attacks, set the limit of pending CSRs to `100 + 2 * number of machines with no node refs`.  Then we always scale with the number of machines we are trying to add to the cluster.

mrunalp:
@enj We discussed a similar solution on a call yesterday :+1:

enxebre:
closing in favour of https://github.com/openshift/cluster-machine-approver/pull/31#issuecomment-510732379. I'll put a new PR for it.

PR #33, opened 2019-07-23
Increase pending csr limit by account for machines

michaelgugino:
Currently, if a user requests a great number of
machines at once, the number of pending CSRs might
increase above the 100 limit, prevent any new
machines from having CSRs approved.

This commit increases the limit to allow CSR
approval during large scale up operations.

https://jira.coreos.com/browse/CLOUD-605


main.go
```diff
@@ -111,20 +111,48 @@ func (c *Controller) handleNewCSR(key string) error {
 		return nil
 	}
 
-	if pending := recentlyPendingCSRs(c.indexer); pending > maxPendingCSRs {
-		klog.Infof("ignoring all CSRs as too many recent pending CSRs seen: %d", pending)
-		return nil
+	machines, err := c.machines.List(metav1.ListOptions{})
```
enxebre: can we get all this to a function e.g `calculateMaxAllowedPendingCSRs()` and unit test it?

main.go
```diff
@@ -111,20 +111,48 @@ func (c *Controller) handleNewCSR(key string) error {
...
 
-	parsedCSR, err := parseCSR(csr)
+	pendingMachinesCount := 0
+	for _, machine := range machines.Items {
+		if machine.Status.NodeRef == nil {
+			pendingMachinesCount++
+		}
+	}
+
+	// Since server-side certs are issued after client side is approved, we
```
enxebre: Based on this statement a machine can only have one pending CSR at a time, wouldn't then `maxPendingCSRs + pendingMachinesCount` be enough, why account for approved ones?

If were possible for machine to have more than one at a time couldn't we simplify by `maxPendingCSRs + 2*pendingMachinesCount`?

@enj @sjenning @sttts PTAL

michaelgugino: No.  If we're treating a machine without a nodeRef as a 'pending machine', that only works for the first CSR.  Once that is approved, the nodelink controller will fill in the noderef before the server-side csr is generated.  The number of 'pending machines' would go to zero before the server certs are approved, so the equation becomes 100 + 2* 0.  I put more details of what's happening in the jira ticket.

main.go
```diff
@@ -116,36 +117,19 @@ func (c *Controller) handleNewCSR(key string) error {
...
-	if approvedCount > len(machines.Items) {
-		maxPendingTotal = len(machines.Items) + maxPendingCSRs
+
+	mpt := maxPendingTotal(csrsList.Items, machines.Items, maxPendingOver)
+	if err != nil {
+		return err
 	}
 
-	if pending := recentlyPendingCSRs(c.indexer); pending > maxPendingTotal {
+	if pending := recentlyPendingCSRs(c.indexer); pending > mpt {
```
frobware: I think `maxPendingTotal` reads a lot better, both now and in the future.

frobware: Oh, i see it became a func. So how about `pending`?

main.go
```diff
@@ -111,20 +112,31 @@ func (c *Controller) handleNewCSR(key string) error {
...
 
-	parsedCSR, err := parseCSR(csr)
+	csrsList, err := c.csrs.List(metav1.ListOptions{})
 	if err != nil {
-		klog.Infof("error parsing request CSR: %v", err)
+		return fmt.Errorf("failed to list all csrs: %v", err)
+	}
+
+	maxPending := maxPendingTotal(csrsList.Items, machines.Items, maxPendingOver)
+	if err != nil {
```
enj: copy paste?

main.go
```diff
@@ -149,6 +161,33 @@ func (c *Controller) handleNewCSR(key string) error {
 	return nil
 }
 
+func maxPendingTotal(csrs []certificatesv1beta1.CertificateSigningRequest, machines []machinev1beta1.Machine, maxPendingCSRs int) int {
```
enj: I would expect:

`100 + 2*machines with no node ref + 1*machines with node ref`

i.e. `100 (buffer) + client+serving for new nodes + just serving for old nodes`

main.go
```diff
@@ -111,20 +112,31 @@ func (c *Controller) handleNewCSR(key string) error {
...
+
+	maxPending := maxPendingTotal(csrsList.Items, machines.Items, maxPendingOver)
+	if err != nil {
+		return err
+	}
+
+	if pending := recentlyPendingCSRs(c.indexer); pending > maxPending {
+		klog.Error("Difference between pending CSRs and machines > 100", pending)
+		klog.Errorf("Pending CSRs: %d; Max pending allowed: %d", pending, maxPending)
 		return nil
```
enj: I think it was a mistake for me to `return nil` here in the original implementation - let us return an error to force a few retries.  That way if the indexer cache is lagging behind, we do not give up too early.

main.go
michaelgugino: 2*(machine without node ref) doesn't fit because the second CSR is only generated after the first is approved.

I don't recall what the certificate rotation (if any) flow is.  That might generate both at once which would break us here.  There's also an edge case that if the csr approver is scaled down for some non-trivial amount of time, the kubelet will endless request CSRs every so often, but I'm not sure there's much we can do about this case in particular here.

All in all, I think 100 + 2*machines should probably be the logic in case there's a large inflow of renewals that need signing, there shouldn't be more than 2 CSRs per node.

enj: How about a slightly different approach:

1. Parse CSR
1. If CSR is not for node client cert, do no max pending check (meaning we do not care about having a million pending serving cert CSRs at once)
1. If CSR is for node client cert
    1. check to see how many machines have missing node refs, add `10`, store as `maxPending`
    1. check to see how many node client certs CSRs are pending, must not be greater than `maxPending`

michaelgugino: @enj given this is mostly a bug fix and I'm assuming we want to backport to 4.1, I'd prefer to keep it simple for now.

enj: > @enj given this is mostly a bug fix and I'm assuming we want to backport to 4.1, I'd prefer to keep it simple for now.

The approach in master is simple and has proven to be incorrect.  I find this to also be incorrect, just less so.  Please take the time to do it correctly.  We are not rushing to get this out the door like 4.1.

michaelgugino: @enj I did take the time to do it correctly.  If you'll notice, the original suggestion on how to implement it was wrong.  I don't want to add more complexity to this, if you want to do it another way, feel free to implement it.

PR #34, opened 2019-07-24
Metrics support

vikaschoudhary16:
This PR is:

Adding basic prometheus metrics support to report pending CSR requests
Adding logic to start the server which prometheus servers can use to scrape the metrics
Yamls for the objects required for integrating with openshift-monitoring
## Approach for https support:
[kube-rbac-proxy](https://github.com/openshift/kube-rbac-proxy/tree/master/examples/rewrites#rewriting-subjectaccessreviews-example) container is being run alongside operator container which runs metrics server in the same pod. kube-rbac -proxy validates authz and authn on behalf of the metrics server and if success, proxies the incoming https request to the metrics server(which is listening on 127.0.0.1 and http port 9191).
### How authn and authz is being validated by kube-rbac-proxy
It uses `TokenReviews()` and `SubjectAccessReviews()` apis internally for validation of token authentication and authorization respectively. These apis are used against a stub resource, `namespace/metrics` which is passed as config to the kube-rbac-proxy
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kube-rbac-proxy
  namespace: openshift-cluster-machine-approver
data:
  config-file.yaml: |+
    authorization:
      resourceAttributes:
        apiVersion: v1
        resource: namespace
        subresource: metrics
        namespace: openshift-cluster-machine-approver
 ```
[RBAC permissions](https://github.com/openshift/machine-api-operator/pull/368/commits/cfacd9dc1d5666dd1958dcd3fa6f32c78fd7d7e1#diff-2043d8125934ffff48ca0cb4d630543bR329-R334) for this rule are defined same as that we expect for metrics endpoint.
So when a https request is received for metrics endpoint, token and rbac of the request user is validated against this stub resource. If it passes, proxied to the metrics endpoint.

### CODE REVIEW

**Comment by mrunalp** [comment]
Is there anything we can expose or reuse or add to kube-state-metrics? https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/certificatesigningrequest.go

**Comment by vikaschoudhary16** [comment]
> Is there anything we can expose or reuse from or add to kube-state-metrics? https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/certificatesigningrequest.go

kube-state-metrics are related to CSR objects at apiserver. While these metrics are about internal state of csr approver. At the moment,  not able to think of any useful alertrule that we can create by combination of metrics here and the ones from the kube-state-metrics.

Btw, i am unable to see CSR related kube-state-metrics  eg, `kube_certificatesigningrequest_condition` in the drop down list of metrics at prometheus dashboard of a newly created openshift cluster.

File: main.go [hunk 1]
```diff
@@ -90,35 +98,35 @@ func (c *Controller) processNextItem() bool {
 func (c *Controller) handleNewCSR(key string) error {
 	obj, exists, err := c.indexer.GetByKey(key)
 	if err != nil {
-		klog.Errorf("Fetching object with key %s from store failed with %v", key, err)
+		glog.Errorf("Fetching object with key %s from store failed with %v", key, err)
```
enxebre: Why are we moving to glog? if there were a reason why in this commit? Can we please use the imperative mood in the commit i.e from "Metrics support" to "Add metrics support"

vikaschoudhary16: Currently this packahge is usig both glog and klog, https://github.com/openshift/cluster-machine-approver/blob/master/config.go#L7
While debuggin i just made klog logs were not getting printed because klog.Init(nil) is missing so just converted all to glog. Agree that this should be handled in a seperate PR.

metrics.go
```diff
@@ -0,0 +1,44 @@
...
+func (mc MetricsCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- CurrentPendingCSRCountDesc
+	ch <- MaxPendingCSRDesc
+}
+
+// Collect implements the prometheus.Collector interface.
+func (mc MetricsCollector) collectMetrics(ch chan<- prometheus.Metric) {
+	pending := recentlyPendingCSRs(mc.cacheIndexer)
+	ch <- prometheus.MustNewConstMetric(CurrentPendingCSRCountDesc, prometheus.GaugeValue, float64(pending))
+	ch <- prometheus.MustNewConstMetric(MaxPendingCSRDesc, prometheus.GaugeValue, float64(maxPendingCSRs))
```
enxebre: Once this is put into a function https://github.com/openshift/cluster-machine-approver/pull/33#discussion_r306681863, It must be used here

metrics.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+var (
+	// CurrentPendingCSRCountDesc is a metric to report count of the pending csr in the cluster
+	CurrentPendingCSRCountDesc = prometheus.NewDesc("csr_approver_current_pending_csr_count", "Count of pending CSRs at the cluster level", nil, nil)
```
enxebre: this should be mapi namespaced

metrics.go
```diff
@@ -0,0 +1,44 @@
...
+	// MaxPendingCSRDesc is a metric to report threshold value of the pending csr beyond which csr will be ignored
+	MaxPendingCSRDesc = prometheus.NewDesc("csr_approver_max_pending_csr_count", "Threshold value of the pending CSRs beyond which any new CSR requests will be ignored ", nil, nil)
```
enxebre: this should be mapi namespaced

enxebre: this should be mapi namespaced

enxebre:
@michaelgugino @simonpasquier PTAL

manifests/04-deployment.yaml
```diff
@@ -41,6 +41,11 @@ spec:
           value: "6443"
         - name: KUBERNETES_SERVICE_HOST
           value: "127.0.0.1"
+        - name: METRICS_PORT
+          value: "9090"
```
simonpasquier: Not a hard call but 9090 is the usual port for the Prometheus server itself so it feels a bit confusing here.

metrics.go
simonpasquier: Being used for a gauge, the `_count` suffix should be removed.

metrics.go
simonpasquier: Being used for a gauge, the `_count` suffix should be removed.

enj:
Metrics must require auth to be scraped.

vikaschoudhary16:
> Metrics must require auth to be scraped.

Can you please explain more on what is missing. We were following this guide for integration, https://docs.google.com/document/d/1rCKAYTrYMESjJDyJ0KvNap05NNukmNXVVi6MShISnOw/edit#heading=h.sndx01stw9z9:

vikaschoudhary16:
@enj PTAL now!

manifests/0000_90_cluster-machine-approver_04_alertrules.yaml
```diff
@@ -0,0 +1,20 @@
...
+    role: alert-rules
+  name: machineapprover-rules
+  namespace: openshift-cluster-machine-approver
+spec:
+  groups:
+    - name: general.rules
+      rules:
+        - alert: MachineApproverMaxPendingCSRsReached
+          expr: |
+             mapi_current_pending_csr_count > mapi_max_pending_csr_count
```
enxebre: shouldn't this match the names in https://github.com/openshift/cluster-machine-approver/pull/34/files#diff-c47669012d7664f9cbee69c67bf8bedfR11

metrics.go
enxebre: this seems incorrect. This should be tracking `getMaxPending`

enxebre:
/hold
holding as `approve cancel` is not removing approval label. Please see:
https://github.com/openshift/cluster-machine-approver/pull/34#discussion_r321208174
https://github.com/openshift/cluster-machine-approver/pull/34#discussion_r321207371

enxebre:
@michaelgugino PTAL

main.go
```diff
@@ -117,9 +125,9 @@ func (c *Controller) handleNewCSR(key string) error {
 		return fmt.Errorf("failed to list machines: %v", err)
 	}
 
-	maxPending := getMaxPending(machines.Items)
-	if pending := recentlyPendingCSRs(c.indexer); pending > maxPending {
-		klog.Errorf("Pending CSRs: %d; Max pending allowed: %d. Difference between pending CSRs and machines > %v. Ignoring all CSRs as too many recent pending CSRs seen", pending, maxPending, maxPendingCSRs)
+	maxPendingCSRs = getMaxPending(machines.Items)
```
simonpasquier: There's a potential data race since `maxPendingCSRs` is read during the metrics scrape which happens asynchronously. You need [SwapInt64](https://golang.org/pkg/sync/atomic/#SwapInt64)/[LoadInt64](https://golang.org/pkg/sync/atomic/#LoadInt64) (or variants) here.

enxebre:
closing in favour of https://github.com/openshift/cluster-machine-approver/pull/50

PR #35, opened 2019-07-25
Include alternate internal DNS in address check

csrwng:
In GCP, instances can use a different FQDN depending on when the
organization or project was created.
See: https://cloud.google.com/compute/docs/internal-dns#instance-fully-qualified-domain-names
This change enables the approver to check for either internal DNS as
well as alternate internal DNS.

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -18,6 +18,7 @@ import (
 )
 
 const (
+	altInternalDNS = "AltInternalDNS"
 	nodeUser       = "system:node"
 	nodeGroup      = "system:nodes"
 	nodeUserPrefix = nodeUser + ":"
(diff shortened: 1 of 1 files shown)


csrwng:
The GCP machine controller creates machines with 2 potential internal DNS addresses:
https://github.com/openshift/cluster-api-provider-gcp/blob/master/pkg/cloud/gcp/actuators/machine/reconciler.go#L165-L177
Which one is used depends on when your org/project was enabled for GCP compute.

csr_check.go
```diff
@@ -151,7 +152,7 @@ func authorizeCSR(config ClusterMachineApproverConfig, machines []v1beta1.Machin
 		var foundSan bool
 		for _, addr := range targetMachine.Status.Addresses {
 			switch addr.Type {
-			case corev1.NodeInternalDNS, corev1.NodeExternalDNS, corev1.NodeHostName:
+			case corev1.NodeInternalDNS, corev1.NodeExternalDNS, corev1.NodeHostName, altInternalDNS:
```
russellb: Instead of using the custom `AltInternalDNS`, why not have 2 entries with a type of `corev1.NodeInternalDNS` and then make sure this code doesn't stop checking on the first one?

csrwng: That sounds good. I don't like introducing yet another key.

csrwng: So it looks like it already does that ... will close this PR and make the change over there. Thx!

Issue #36, opened 2019-07-26
Timeout not long enough


zeenix:
Ever since d7e76c48c4f29e7bef109e6aed5d400c0e6cbd4b, the CSR on not-so-fast hosts (e.g nested-virt) [don't get auto-approved any more](https://github.com/openshift/installer/issues/1893) and people have to manually approve them for cluster to initialize. After an enlightening discussion with the commit author @enj, I'm not sure what can be done in the short term for this but we have to fix this sooner or later somehow.

zeenix:
Independent report of this on baremetal: https://bugzilla.redhat.com/show_bug.cgi?id=1723955

cgwalters:
I think if we tied together the MCO and machine API more, we could have the MCO inject a per-machine secret into the cloud-init data provided, tied to the machine resource.  This would give us a way for booted nodes to prove their identity, and CSR approval could be tied to that.  There wouldn't be any need for timeouts (or, they could be way, way longer because we're not worried about machine reuse).

enj:
> I think if we tied together the MCO and machine API more, we could have the MCO inject a per-machine secret into the cloud-init data provided, tied to the machine resource. This would give us a way for booted nodes to prove their identity, and CSR approval could be tied to that. There wouldn't be any need for timeouts (or, they could be way, way longer because we're not worried about machine reuse).

+100

The current approach is just meant as a series of probabilities that reduce the impact of lost bootstrap credentials.  With proper cryptographic node attestation, the timing and abuse checks simply disappear as we can easily determine the identity of a node.

russellb:
The approach of using per-machine secrets sounds great.

> Independent report of this on baremetal: https://bugzilla.redhat.com/show_bug.cgi?id=1723955

In this issue, it occurs on AWS with a bare metal instance type.  This problem will also come up for the `baremetal` infrastructure platform type, where we're automating the provisioning of bare metal hosts.  It can take more than 10 minutes for a server to boot.  When automating provisioning on prem, we'll have a minimum of two boots to wait for.

cgwalters:
The tricky thing about this is that in a machineAPI managed environment, we control the whole stack and are in a position to do lots of sophisticated things.  But in e.g. a bare metal PXE environment (non-metalkube), we don't really have a programmatic way to inject data into the booted machines.   On the other hand, I think in a "manual bare metal" env you really end up needing/wanting to manually approve CSRs anyways.

russellb:
[...]

PR #37, opened 2019-08-01
Increase machine creation time limit

michaelgugino:
Increase machine creation time limit

GCP sometimes takes longer than 10 minutes to provision
a machine to the point of requesting a CSR.

Increase the limit slightly to allow more approvals.

Fixes: https://github.com/openshift/cluster-machine-approver/issues/36

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -28,7 +28,7 @@ const (
 	nodeBootstrapperUsername = "system:serviceaccount:openshift-machine-config-operator:node-bootstrapper"
 
 	maxMachineClockSkew = 10 * time.Second
-	maxMachineDelta     = 10 * time.Minute
+	maxMachineDelta     = 2 * time.Hour
 )
 
 var nodeBootstrapperGroups = sets.NewString(
--- a/csr_check_test.go
+++ b/csr_check_test.go
@@ -230,7 +230,7 @@ Certificate Request:
         Subject: O = system:nodes, CN = system:node:monkey
 ...
         Requested Extensions:
-            X509v3 Subject Alternative Name: 
+            X509v3 Subject Alternative Name:
                 DNS:banana
 ...
 -----BEGIN CERTIFICATE REQUEST-----
@@ -1903,7 +1903,7 @@ func Test_authorizeCSR(t *testing.T) {
 				},
 				csr: clientGood,
 			},
-			wantErr: "CSR purple creation time 2000-01-01 02:32:00 +0000 UTC not in range (2000-01-01 02:32:50 +0000 UTC, 2000-01-01 02:43:00 +0000 UTC)",
+			wantErr: "CSR purple creation time 2000-01-01 02:32:00 +0000 UTC not in range (2000-01-01 02:32:50 +0000 UTC, 2000-01-01 04:33:00 +0000 UTC)",
 		},
 		{
 			name: "client good but CSR too late",
@@ -1938,7 +1938,7 @@ func Test_authorizeCSR(t *testing.T) {
 				req: &certificatesv1beta1.CertificateSigningRequest{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:              "red",
-						CreationTimestamp: creationTimestamp(15 * time.Minute),
+						CreationTimestamp: creationTimestamp(25 * time.Hour),
 					},
 					Spec: certificatesv1beta1.CertificateSigningRequestSpec{
 						Usages: []certificatesv1beta1.KeyUsage{
@@ -1956,7 +1956,7 @@ func Test_authorizeCSR(t *testing.T) {
 				},
 				csr: clientGood,
 			},
-			wantErr: "CSR red creation time 2000-01-01 02:45:00 +0000 UTC not in range (2000-01-01 02:32:50 +0000 UTC, 2000-01-01 02:43:00 +0000 UTC)",
+			wantErr: "CSR red creation time 2000-01-02 03:30:00 +0000 UTC not in range (2000-01-01 02:32:50 +0000 UTC, 2000-01-01 04:33:00 +0000 UTC)",
 		},
 	}
 	for _, tt := range tests {


csr_check_test.go
```diff
@@ -1903,7 +1903,7 @@ func Test_authorizeCSR(t *testing.T) {
 				},
 				csr: clientGood,
 			},
-			wantErr: "CSR purple creation time 2000-01-01 02:32:00 +0000 UTC not in range (2000-01-01 02:32:50 +0000 UTC, 2000-01-01 02:43:00 +0000 UTC)",
+			wantErr: "CSR purple creation time 2000-01-01 02:32:00 +0000 UTC not in range (2000-01-01 02:32:50 +0000 UTC, 2000-01-02 02:33:00 +0000 UTC)",
```
frobware: If this is bumping by 24h why are the minutes not `43`?

frobware: The delta was previously 10m. Gotcha.

enxebre:
why increasing to 24h? if anything I think we should set to slowest machine time creation

michaelgugino:
> why increasing to 24h? if anything I think we should set to slowest machine time creation

@enxebre what precisely are we trying to prevent with this?  I'd prefer this not silently break people, 24 hours seems reasonable to me.  Maybe 6 hours would be okay too.  I just want it to be well outside our remediation window.

enxebre:
Node names are predictable, we want to keep the window as short as possible i.e what we consider the threshold for the slowest machine creation time (in future we could adjust tracking the time the machine was "provisioned") to reduce the chances of an attacker preloading CSRs and guessing the correct name of a new node. Increasing this way has no benefits and goes against the best effort the approver tries to do.
Let's try to address https://github.com/openshift/cluster-machine-approver/pull/33#discussion_r307436165 and make sure we get https://github.com/openshift/cluster-machine-approver/pull/33 merged before any farther change

csr_check.go
```diff
@@ -28,7 +28,7 @@ const (
 	nodeBootstrapperUsername = "system:serviceaccount:openshift-machine-config-operator:node-bootstrapper"
 
 	maxMachineClockSkew = 10 * time.Second
-	maxMachineDelta     = 10 * time.Minute
+	maxMachineDelta     = 20 * time.Minute
```
enxebre: 10 min is currently making CSRs in GCP to occasionally fail. This is likely to because how LB/bootstrap setup works in GCP we need to wait for the bootstrap machine to be torn down and the LB to switch over the final control plane for the ignition secrets to be reachable. See https://github.com/openshift/installer/pull/2050 cc @abhinavdahiya 
@michaelgugino can we update the commit long description to reflect this and add also add comment here?
For baremetal if a big increment is needed we might want to allow it only for baremetal by checking the platform

dhellmann:
For baremetal, we haven't seen timeouts yet but I didn't even realize there was one. I can easily see provisioning a host taking more than 10 minutes between setting up RAID, writing the image to disk, and then booting the host. Some of these boxes take several minutes just to reboot because of all of the hardware checks that are performed.

I'm not sure we will always be able to predict a fixed amount of time, and I can understand wanting this timeout to be as short as possible. Given that it's likely to be platform-specific, I wonder if it's something we should make configurable (at least so the installer binary can set a different value based on the platform, even if we don't tell users how to change that).

mhrivnak:
To quickly expand on baremetal: It could take 10+ minutes just for some servers to boot. As I understand the provisioning process, the OS image that gets written will do an "early pivot"; upon booting, it will retrieve a newer OS image, install that, and then reboot. So we could be looking at 2 reboot cycles, plus the time it takes to write the images, download at least one of the images, etc. A total of 30+ minutes could be normal in some scenarios.

dhellmann:
> To quickly expand on baremetal: It could take 10+ minutes just for some servers to boot. As I understand the provisioning process, the OS image that gets written will do an "early pivot"; upon booting, it will retrieve a newer OS image, install that, and then reboot. So we could be looking at 2 reboot cycles, plus the time it takes to write the images, download at least one of the images, etc. A total of 30+ minutes could be normal in some scenarios.

We might have 3 reboots if we aren't already running the ramdisk image used to write the RHCOS image to disk.

zeenix:
@michaelgugino Could you please add "Fixes #36" to commit log and description here?

enxebre:
I think we should a least document https://github.com/openshift/cluster-machine-approver/pull/37#discussion_r310671173 adjust the timing to that and increase broader only for baremetal.

Before shipping 4.1 it was proven the cluster was vulnerable and it was possible for an untrusted user to escalate privileges, see: https://github.com/openshift/machine-config-operator/issues/731
https://bugzilla.redhat.com/show_bug.cgi?id=1707162.

Because the set of all valid node names can be guessed, an attacker could pre-load all possible CSRs, then trigger autoscaling. The machine approver it's a best effort to help to mitigate this issue. With the above PR and https://github.com/openshift/cluster-machine-approver/pull/33 we are making it more vulnerable

michaelgugino:
> I think we should a least document [#37 (comment)](https://github.com/openshift/cluster-machine-approver/pull/37#discussion_r310671173) adjust the timing to that and increase broader only for baremetal.
> 
> Before shipping 4.1 it was proven the cluster was vulnerable and it was possible for an untrusted user to escalate privileges, see: [openshift/machine-config-operator#731](https://github.com/openshift/machine-config-operator/issues/731)
> https://bugzilla.redhat.com/show_bug.cgi?id=1707162.
> 
> Because the set of all valid node names can be guessed, an attacker could pre-load all possible CSRs, then trigger autoscaling. The machine approver it's a best effort to help to mitigate this issue. With the above PR and #33 we are making it more vulnerable

@enxebre 
You cannot preload the CSRs before triggering autoscaling.  The CSRs have to be newer than the machine-creation timestamp.

If a machine already has a noderef, the client csr will not be approved, so as long as the machines become nodes in a reasonable amount of time, the 2 hours won't matter.

PR #38, opened 2019-08-23
Bug 1737611: Simplified approval flow for renewing serving certs

bison:
This adds a simplified flow for approving renewals of existing kubelet
serving certificates that does not rely directly on details from the
machine-api.  When approving a CSR for a serving certificate, the
machine-approver will attempt to connect to the requesting node's
kubelet on its first advertised address. If successful, the returned
TLS certificate will be used to validate the new CSR.  All of the
following must be true to authorize the new CSR:

 - Existing cert is valid and signed by the current CA.
 - All Subject Alternate Name values are equal.

If the CA cannot be fetched, or no current serving certificate is
found, due to a connection error for example, the previous flow
relying on the machine-api is used as a fallback.


bison:
I'm really unsure about the best way to go about the suggested check that the existing cert "has the same private key as the new CSR" in a way that works for all key types.

Even so, on my test cluster I get new key material for each renewal, which means that test is not valid anyway.

smarterclayton:
After reviewing the private key requirement, I think we can loosen that to "is being requested by an entity that is the current node" - if the node is compromised then anyone can impersonate the node on renewal, and we don't attempt to prevent that today.  The private key is a slightly different way of expressing that - basically instead of you having the same private key as the server CSR, you have a client who via a private key for the client has proven that they are the same node as they were previously.

csr_check.go
```diff
@@ -119,7 +124,14 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
...
+func authorizeCSR(
+	config ClusterMachineApproverConfig,
+	machines []v1beta1.Machine,
+	nodes corev1client.NodeInterface,
+	req *certificatesv1beta1.CertificateSigningRequest,
+	csr *x509.CertificateRequest,
+	ca *x509.CertPool,
+) error {
 	if len(machines) == 0 || req == nil || csr == nil {
 		return fmt.Errorf("Invalid request")
```
enxebre: This check `len(machines) == 0 ` would defeat the intention of the PR.

enxebre:
I'm good with the approach for renewal. (Until there's an approval flow possibly decoupled from the machine api) we need to be careful though and don't deviate the design of this machine approver to fill the UPI lacks/gaps where none of the design assumptions here are guaranteed.

csr_check.go
bison: Good call. I hadn't actually run this outside AWS. Removed that check.

bison:
This seems to be working for me on AWS. It could use unit test coverage and testing on an actual UPI cluster though.

How I'm testing:
  - Add a CVO override and run a custom image, or just run the operator locally.
  - Edit the kube-controller-manager config: `kubectl edit kubecontrollermanager cluster`
  - Set `experimental-cluster-signing-duration` to 10 minutes:
```yaml
    unsupportedConfigOverrides:  
      extendedArguments:                                                                     
        experimental-cluster-signing-duration:
        - 10m
 ```

Now you can scale up a MachineSet. The new node should have its certificates approved, then renewed every few minutes.

**Comment by bison** [comment]
I have one concern here still: What happens when the Kubelet serving CA expires?

It looks like the kube-controller-manager-operator rotates the Kubelet CA every 30 days. The kube-controller-manager will not issue a certificate with a validity past that of its signing CA. So as the expiration of the CA approaches, everything must expire at once, right?

If that's the case, when we go to renew these serving certs, we'll connect to the node and get a certificate, but it will be signed by the old CA, which we don't have any more. Are the previous CA certificates stored anywhere? If so, I guess we could try to verify the presented certificate against all of them.

cc: @enxebre @smarterclayton @mrunalp

**Comment by smarterclayton** [comment]
Renewal on the Kubelet should happen more often than CA expiration.  If
not, that’s def a bug.

On Sep 11, 2019, at 12:09 PM, Brad Ison <notifications@github.com> wrote:

I have one concern here still: What happens when the Kubelet serving CA
expires?

It looks like the kube-controller-manager-operator rotates the Kubelet CA
every 30 days. The kube-controller-manager will not issue a certificate
with a validity past that of its signing CA. So as the expiration of the CA
approaches, everything must expire at once, right?

If that's the case, when we go to renew these serving certs, we'll connect
to the node and get a certificate, but it will be signed by the old CA,
which we don't have any more. Are the previous CA certificates stored
anywhere? If so, I guess we could try to verify the presented certificate
against all of them.

cc: @enxebre <https://github.com/enxebre> @smarterclayton
<https://github.com/smarterclayton> @mrunalp <https://github.com/mrunalp>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/cluster-machine-approver/pull/38?email_source=notifications&email_token=AAI37J4SISJPTGWUIXHOP23QJEJ3JA5CNFSM4IPAXFT2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD6PAXJI#issuecomment-530451365>,
or mute the thread

File: csr_check.go [hunk 2]
```diff
@@ -236,6 +269,46 @@ func authorizeNodeClientCSR(config ClusterMachineApproverConfig, machines []v1be
...
+	}
+
+	// Check that the serving cert is signed by the given CA, is not expired,
+	// and is otherwise valid.
+	if _, err := currentCert.Verify(x509.VerifyOptions{Roots: ca}); err != nil {
+		return err
+	}
+
+	// Check that the CN is correct on the current cert.
+	if currentCert.Subject.CommonName != fmt.Sprintf("system:node:%s", nodeName) {
```
enxebre: could use for `system:node:` constant here

smarterclayton:
You should be able to use cluster-bot to launch vsphere and then try it

PR #42, opened 2019-08-30
process pending CSRs

rphillips:
The bootstrap node approves pending CSRs while the cluster is installing. Once the bootstrap node is destroyed the auto approve script on the bootstrap node stops. The gap of time between the bootstrap node being destroyed and the auto approver starting requires a loop through the CSRs to attempt to approve the pending ones.

### DIFF
--- a/main.go
+++ b/main.go
@@ -102,6 +102,10 @@ func (c *Controller) handleNewCSR(key string) error {
 
 	// do not mutate informer cache
 	csr := obj.(*certificatesv1beta1.CertificateSigningRequest).DeepCopy()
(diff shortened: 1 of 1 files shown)


mrunalp:
@bison @enj ptal
cc: @sjenning

wking:
Will this apply to upgrades or other hiccups which lead to machine-approver downtime as well?  The new approver comes up, attaches its listener, and then checks for anything outstanding from before the listener was setup?

rphillips:
@wking in this patch the CSRs are checked before the controller loop starts.

PR #43, opened 2019-09-04
Increase pending allowed

enxebre:
Re open https://github.com/openshift/cluster-machine-approver/pull/33 squashed in one commit

### DIFF
--- a/main.go
+++ b/main.go
@@ -118,10 +118,9 @@ func (c *Controller) handleNewCSR(key string) error {
 	}
 
 	maxPending := getMaxPending(machines.Items)
-
 	if pending := recentlyPendingCSRs(c.indexer); pending > maxPending {
-		klog.Errorf("Pending CSRs: %d; Max pending allowed: %d", pending, maxPending)
-		return fmt.Errorf("Difference between pending CSRs and machines > 100: %v", pending)
+		klog.Errorf("Pending CSRs: %d; Max pending allowed: %d. Difference between pending CSRs and machines > %v. Ignoring all CSRs as too many recent pending CSRs seen", pending, maxPending, maxPendingCSRs)
+		return nil
 	}
 
 	parsedCSR, err := parseCSR(csr)
--- a/main.go
+++ b/main.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
+	"github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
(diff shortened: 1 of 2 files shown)


enxebre:
cc @michaelgugino @ingvagabund @mrunalp

main.go
```diff
@@ -111,20 +111,48 @@ func (c *Controller) handleNewCSR(key string) error {
...
+		klog.Errorf("Pending CSRs: %d; Pending Machines %d", pending, pendingMachinesCount)
 		return nil
 	}
 
-	machines, err := c.machines.List(metav1.ListOptions{})
+	parsedCSR, err := parseCSR(csr)
 	if err != nil {
-		return fmt.Errorf("failed to list machines: %v", err)
+		klog.Infof("error parsing request CSR: %v", err)
+		return nil
```
frobware: Why does this no longer return an error?

main.go
```diff
@@ -111,20 +111,48 @@ func (c *Controller) handleNewCSR(key string) error {
...
+	}
+	maxPendingTotal := maxPendingCSRs + pendingMachinesCount + approvedCount
+	// If for some reason we see a really high approved count, just set
+	// absolute max to total machines + maxPendingCSRs
+	if approvedCount > len(machines.Items) {
+		maxPendingTotal = len(machines.Items) + maxPendingCSRs
+	}
+
+	if pending := recentlyPendingCSRs(c.indexer); pending > maxPendingTotal {
+		klog.Error("Difference between pending CSRs and machines > 100", pending)
```
frobware: We should avoid the hardcoded "> 100"; format appropriately.

main.go
enxebre: this behaviour is un changed. If failed to parse the queue forget this key so AddRateLimited is not applied for future processing of updates https://github.com/openshift/cluster-machine-approver/blob/master/main.go#L153

main.go
```diff
@@ -111,9 +112,16 @@ func (c *Controller) handleNewCSR(key string) error {
...
+	machines, err := c.machines.List(metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list machines: %v", err)
+	}
+
+	maxPending := getMaxPending(machines.Items)
+
+	if pending := recentlyPendingCSRs(c.indexer); pending > maxPending {
+		klog.Errorf("Pending CSRs: %d; Max pending allowed: %d", pending, maxPending)
+		return fmt.Errorf("Difference between pending CSRs and machines > 100: %v", pending)
```
enxebre: @michaelgugino we should preserve the return nil here so upcoming CSRs are ignored

PR #44, opened 2019-09-04
Add consumer given values to logs

enxebre:
Add consumer given values to logs

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -39,7 +39,7 @@ var nodeBootstrapperGroups = sets.NewString(
 
 func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr *x509.CertificateRequest) (string, error) {
 	if !strings.HasPrefix(req.Spec.Username, nodeUserPrefix) {
-		return "", fmt.Errorf("Doesn't match expected prefix")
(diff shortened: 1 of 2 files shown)


csr_check.go
```diff
@@ -39,7 +39,7 @@ var nodeBootstrapperGroups = sets.NewString(
 
 func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr *x509.CertificateRequest) (string, error) {
 	if !strings.HasPrefix(req.Spec.Username, nodeUserPrefix) {
-		return "", fmt.Errorf("Doesn't match expected prefix")
+		return "", fmt.Errorf("%q doesn't match expected prefix: %q", req.Spec.Username, nodeUserPrefix)
```
ingvagabund: @cgwalters did you omit the values on purposes due to limiting the data that gets logged due to security reasons? Or, in the time you added better logging message [1] it was not necessary?

[1] https://github.com/openshift/cluster-machine-approver/commit/374b20416ec80f8a05a5b178d6b2313553c04a0f#diff-a482a1a664767249971b0b71ca9baaf0

cgwalters: Nope, I didn't think about it too much.  I don't see a reason not to log more data, only kubeadmin should have access to the CSR logs right?

ingvagabund: I assume that's true in production. Thanks.

PR #46, opened 2019-09-05
Refactor code within authorizeCSR into meaningful named funcs

enxebre:
Refactor code within authorizeCSR into meaningful named funcs


csr_check.go
```diff
@@ -128,12 +138,21 @@ func authorizeCSR(config ClusterMachineApproverConfig, machines []v1beta1.Machin
...
 	}
 
-	// node serving cert validation after this point
+	if isNodeServingCert(req, csr) {
+		return authorizeNodeServingCSR(machines, req, csr)
+	}
+
+	klog.Infof("CSR %q is not a client/serving CSR", req.Name)
+	// we don't want to requeue if none of the above was met so we return nil.
+	return nil
```
russellb: The PR describes the changes as a refactor, but it looks like there's a change in behavior here.  Previously, an error would have been returned if it wasn't a client CSR, and didn't match the expected format of a server CSR.  The changes seem fine, but would you like to update the description?

ingvagabund:
@enxebre is this a bug fix? Or, just code improvement?

csr_check.go
enxebre: I'm actually keeping current error returning to keep unit test coverage and minimise scope of the PR.
I'll do farther refinements as small follow ups

csr_check.go
```diff
@@ -37,33 +37,33 @@ var nodeBootstrapperGroups = sets.NewString(
...
-		return "", fmt.Errorf("Empty name")
+		return false, fmt.Errorf("%q has empty node name", req.Spec.Username)
 	}
 
 	// Check groups, we need at least:
 	// - system:nodes
 	// - system:authenticated
 	if len(req.Spec.Groups) < 2 {
-		return "", fmt.Errorf("Too few groups")
+		return false, fmt.Errorf("Too few groups %v", req.Spec.Groups)
```
frobware: It would be better to convey many were needed - or that there are `< 2`.

csr_check.go
```diff
@@ -128,12 +128,20 @@ func authorizeCSR(config ClusterMachineApproverConfig, machines []v1beta1.Machin
 		return authorizeNodeClientCSR(config, machines, nodes, req, csr)
 	}
 
-	// node serving cert validation after this point
+	isNodeServingCert, err := isNodeServingCert(req, csr)
+	if isNodeServingCert {
```
frobware: Shouldn't this check `err` first? If the implementation changed to return `true, a-real-error` this would unexpectedly pass.

csr_check.go
```diff
@@ -128,12 +128,20 @@ func authorizeCSR(config ClusterMachineApproverConfig, machines []v1beta1.Machin
...
+	return err
+}
 
-	nodeAsking, err := validateCSRContents(req, csr)
-	if err != nil {
-		return err
+func authorizeNodeServingCSR(machines []v1beta1.Machine, req *certificatesv1beta1.CertificateSigningRequest, csr *x509.CertificateRequest) error {
+	nodeAsking := strings.TrimPrefix(req.Spec.Username, nodeUserPrefix)
+	if len(nodeAsking) == 0 {
+		return fmt.Errorf("Empty name")
```
frobware: Empty `username`. Might help convey what's wrong.

csr_check_test.go
```diff
@@ -455,7 +455,7 @@ func Test_authorizeCSR(t *testing.T) {
 				},
 				csr: goodCSR,
 			},
-			wantErr: "Empty name",
+			wantErr: "\"system:node:\" has empty node name",
```
frobware: Change to a string literal - would be easier to read.

PR #48, opened 2019-09-25
Bug 1755608: Always fall back to machine-api check for serving cert

michaelgugino:
This commit ensures we always fall back to machine-api
serving cert in case we hit an error validating the existing
serving cert compared to the new one.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1755608

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -156,7 +156,12 @@ func authorizeCSR(
 		servingCert, err := getServingCert(nodes, nodeAsking, ca)
 		if err == nil && servingCert != nil {
 			klog.Infof("authorizing serving cert renewal for %s", nodeAsking)
-			return authorizeServingRenewal(nodeAsking, csr, servingCert, ca)
+			if authorizeErr := authorizeServingRenewal(nodeAsking, csr, servingCert, ca); authorizeErr != nil {
+				// Don't return here so we can fallback to machine-api check.
+				klog.Errorf("%v: unable to validate existing serving cert for node %v: %v", req.Name, nodeAsking, authorizeErr)
+			} else {
+				return nil
+			}
 		}
 
 		if err != nil {
@@ -165,7 +170,7 @@ func authorizeCSR(
 	}
 
 	// Fall back to the original machine-api based authorization scheme.
-	klog.Infof("No existing serving certificate found for %s", nodeAsking)
+	klog.Infof("%v: Falling back to machine-api based check %s", req.Name, nodeAsking)
 
 	// Check that we have a registered node with the request name
 	targetMachine, ok := findMatchingMachineFromNodeRef(nodeAsking, machines)


csr_check.go
```diff
@@ -156,7 +156,12 @@ func authorizeCSR(
...
-			return authorizeServingRenewal(nodeAsking, csr, servingCert, ca)
+			authorizeErr := authorizeServingRenewal(nodeAsking, csr, servingCert, ca)
+			if authorizeErr == nil {
+				return nil
+			}
+			// Don't return here so we can fallback to machine-api check.
+			klog.Errorf("%v: unable to validated existing serving cert for node %v: %v", req.Name, nodeAsking, authorizeErr)
 		}
 
 		if err != nil {
```
ingvagabund: Is it possible to get `err != nil` but `servingCert == nil`?

csr_check.go
```diff
@@ -156,7 +156,12 @@ func authorizeCSR(
 		servingCert, err := getServingCert(nodes, nodeAsking, ca)
 		if err == nil && servingCert != nil {
 			klog.Infof("authorizing serving cert renewal for %s", nodeAsking)
-			return authorizeServingRenewal(nodeAsking, csr, servingCert, ca)
+			authorizeErr := authorizeServingRenewal(nodeAsking, csr, servingCert, ca)
```
frobware: nit: I would inline this as we don't care for the error var.

csr_check.go
```diff
@@ -156,7 +156,12 @@ func authorizeCSR(
...
+			if authorizeErr == nil {
+				return nil
+			}
+			// Don't return here so we can fallback to machine-api check.
+			klog.Errorf("%v: unable to validated existing serving cert for node %v: %v", req.Name, nodeAsking, authorizeErr)
```
frobware: nit: s/validated/validate

csr_check.go
```diff
@@ -156,7 +156,12 @@ func authorizeCSR(
 		servingCert, err := getServingCert(nodes, nodeAsking, ca)
 		if err == nil && servingCert != nil {
 			klog.Infof("authorizing serving cert renewal for %s", nodeAsking)
-			return authorizeServingRenewal(nodeAsking, csr, servingCert, ca)
+			if authorizeErr := authorizeServingRenewal(nodeAsking, csr, servingCert, ca); authorizeErr != nil {
```
enxebre: Can we include  a unit test for this case?

enxebre:
/lgtm
Change lgtm. Not sure this will fix the bugzilla issue though, interestingly `oc get csr` gives a list of pending CSR requested by `system:serviceaccount:openshift-machine-config-operator:node-bootstrapper` which should be only the case for the bootstrapping client certs. So one guess is that something getting corrupted at some point for a node might be generating indefinite CSRs causing the machine approver to reach the short circuiting limit, hence preventing it from handling the renewal properly.

michaelgugino:
> /lgtm
> Change lgtm. Not sure this will fix the bugzilla issue though, interestingly `oc get csr` gives a list of pending CSR requested by `system:serviceaccount:openshift-machine-config-operator:node-bootstrapper` which should be only the case for the bootstrapping client certs. So one guess is that something getting corrupted at some point for a node might be generating indefinite CSRs causing the machine approver to reach the short circuiting limit, hence preventing it from handling the renewal properly.

If this were the case, the nodes would not show up in the output of 'oc get nodes'.  If these are client renewals that never got approved, possibly the server renewals were never generated.  If that's the case, then the problem is with the controller-manager, not the autoapprover.

bison:
I'm not against merging this, but I agree that it likely won't solve the issue in the linked Bugzilla ticket. What we're seeing there (when we've been able to get logs...) is that serving cert renewals are successful, but are happening more often than we'd expect. Then we eventually see new client CSRs from the bootstrap user which are rejected because a node object already exists.

It looks to me like this may actually be a problem with the kube-controller-manager holding onto the initial 24 hour CA generated by the installer for longer than it should. When that expires we're not able to generate renew client certs and the cluster goes unhealthy.

bison:
/close

Closing in favor of: https://github.com/openshift/cluster-machine-approver/pull/57

PR #49, opened 2019-09-25
Bug 1755469: Always fall back to machine-api check for serving cert

michaelgugino:
This commit ensures we always fall back to machine-api
serving cert in case we hit an error validating the existing
serving cert compared to the new one.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1755469


mrunalp:
@enxebre @bison ptal

bison:
Just a note that most discussion is happening in Bugzilla and in the counterpart to this PR: https://github.com/openshift/cluster-machine-approver/pull/48

bison:
/close

Closing in favor of: https://github.com/openshift/cluster-machine-approver/pull/57 (and its cherry-picks)

PR #52, opened 2019-10-31
Fix ClusterMachineApproverDown alert

spangenberg:


enxebre:
can you please enable the `e2e-aws` suite here first so this will block unless working as expected?

manifests/03-metrics-service.yaml
```diff
@@ -12,8 +12,9 @@ spec:
   clusterIP: None
   ports:
   - name: https
-    port: 9192
-    targetPort: https
+    protocol: TCP
```
enxebre: why?

enxebre: metrics are already being reported https://xlugntax-promecieus.svc.ci.openshift.org/graph?g0.range_input=1w&g0.stacked=1&g0.moment_input=2019-10-15%2014%3A55%3A54&g0.expr=mapi_max_pending_csr&g0.tab=0

spangenberg: Thanks, confused the named port.

PR #59, opened 2019-12-09
[release-4.2] Bug 1781160: Fall back to machine-api check if certificate check fails

bison:
This falls back to the original machine-api based checks for serving
certificate renewals if the checks based on retrieving the current
serving certificate fail. This is needed in cases where network
interfaces are added to a machine after its initial creation, because
the SAN list in the new CSR will contain new addresses and no longer
match the current certificate exactly.

This requires the machine-controller to track all addresses on a
machine, which was recently fixed for AWS: https://github.com/openshift/cluster-api-provider-aws/pull/279

Cherry-picks for other relases:
https://github.com/openshift/cluster-machine-approver/pull/57
https://github.com/openshift/cluster-machine-approver/pull/58

This also cherry-picks: https://github.com/openshift/cluster-machine-approver/pull/51

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -157,17 +157,27 @@ func authorizeCSR(
 	if ca != nil {
 		servingCert, err := getServingCert(nodes, nodeAsking, ca)
 		if err == nil && servingCert != nil {
-			klog.Infof("authorizing serving cert renewal for %s", nodeAsking)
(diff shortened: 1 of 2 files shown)


bison:
/hold cancel

There's no reason to hold this???

PR #61, opened 2019-12-12
Fix IPv6 connections in getServingCert

bison:
Connections to IPv6 addresses were failing in `getServingCert` because
the address literal was not properly enclosed in square brackets.
This uses `net.JoinHostPort`, which handles this, in place of a simple
`fmt.Sprintf` call.

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


csr_check.go
```diff
@@ -428,11 +428,25 @@ func getServingCert(nodes corev1client.NodeInterface, nodeName string, ca *x509.
...
+
+				continue
+			}
+
+			// IPv6 addresses must be enclosed in square brackets.
+			if ip.To4() == nil {
+				return fmt.Sprintf("[%s]", ip.String()), nil
+			}
+
+			return ip.String(), nil
```
russellb: Instead of doing all this, you can just change line 410 from:
```
	kubelet := fmt.Sprintf("%s:%d", host, port)
```

to something like:
```
kubelet := net.JoinHostPort(host, port)
```

except you have to convert port to a string first, i think

bison: True. Fixed.

russellb:
Since you have a hold on this while you test it to verify, the changes /lgtm

bison:
/hold cancel

I've tested this on an AWS IPv6 cluster and it seems to be working.

PR #64, opened 2020-03-02
Bug 1809329: ROKS - remove ClusterMachineApproverDown prometheus rule

csrwng:
Because the cluster machine approver is not deployed in a ROKS cluster, the alert that says that it's down should not be firing.


derekwaynecarr:
ci failures look unrelated.

/approve
/lgtm

PR #68, opened 2020-03-27
Add RBAC signer rule for 1.18

marun:
Successful TLS kubelet bootstrap for the [1.18 rebase](https://github.com/openshift/origin/pull/24719) requires that the approver have new rbac permissions as per [recent changes to the upstream approver](https://github.com/kubernetes/kubernetes/pull/88246).

### DIFF
--- a/manifests/01-rbac.yaml
+++ b/manifests/01-rbac.yaml
@@ -57,6 +57,15 @@ rules:
   verbs:
   - update
 - apiGroups:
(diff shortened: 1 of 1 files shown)


ingvagabund:
@enxebre we need this to be merged to unblock 1.18 rebase
/lgtm
In case these causes trouble feel free to alter the changes.

PR #71, opened 2020-04-22
Bug 1794678: Add clusteroperator status

enxebre:
This adds a controller to manage the clusterOperator status for the cluster-machine-approver which is currently missing.

Reconciliation is driven by a `status.VersionGetter` and a single object informer watcher.

"Available" is always reported atm but this enables the base to set degraded in desired scenarios in the near future e.g too many pending CSRs.

This also enables integration with must gather.


status_suite_test.go
```diff
@@ -0,0 +1,67 @@
...
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("clusteroperator.crd.yaml")},
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
+
+}, 60)
```
JoelSpeed, line 62: Don't need this 60 here, it would add a timeout to the before suite in case any async blocks but there's no async here
```suggestion
})
```

status_test.go
```diff
@@ -0,0 +1,193 @@
...
+				if v1helpers.IsStatusConditionTrue(co.Status.Conditions, osconfigv1.OperatorProgressing) {
+					return false
+				}
+
+				// check related objects.
+				if !equality.Semantic.DeepEqual(co.Status.RelatedObjects, relatedObjects) {
+					return false
+				}
+				return true
+			}, timeout).Should(BeTrue())
```
JoelSpeed, line 129: We only need to check that one sync has been successful right? I wonder if this might be simpler if we have a smaller eventually then multiple assertions after, this way around the assertions should be easier to interpret and work out which part of the code failed

```suggestion
			var co *osconfigv1.ClusterOperator
			Eventually(func() (bool, err) {
				var err error
				co, err = osClient.ConfigV1().ClusterOperators().Get(context.Background(), clusterOperatorName, metav1.GetOptions{})
				if err != nil {
					return false, err
				}
				// Successful sync means CO exists and the status is not empty
				return co.Status != osconfigv1.ClusterStatusOperator{}, nil
			}, timeout).Should(BeTrue())
			
			// check version.
			Expect(co.Status.Versions).To(HaveLen(1))
			Expect(co.Status.Versions[0].Name).To(Equal(operatorVersionKey))
			Expect(co.Status.Versions[0].Version).To(Equal(expectedVersion))
			
			// check conditions.
			Expect(v1helpers.IsStatusConditionTrue(co.Status.Conditions, osconfigv1.OperatorAvailable)).To(BeTrue())
			Expect(v1helpers.IsStatusConditionTrue(co.Status.Conditions, osconfigv1.OperatorUpgradeable)).To(BeTrue())
			Expect(v1helpers.IsStatusConditionFalse(co.Status.Conditions, osconfigv1.OperatorDegraded)).To(BeTrue())
			Expect(v1helpers.IsStatusConditionFalse(co.Status.Conditions, osconfigv1.OperatorProgressing)).To(BeTrue())
			
			// check related objects.
			Expect(co.Status.RelatedObjects).To(Equal(relatedObjects))
```

enxebre: mm I'd like that but I don't think we can though. We first create the object, then reconcile status towards expected relatedObjects, then reconcile status towards expected conditions.
There's scenarios where the object and status exists so eventually would succeed but we might be still reconciling status, right?

JoelSpeed: Ahh yes, I see we are making two calls to update status, this wont work right now, I'm not too fussed either way tbh, I don't think changing the code to fit the tests makes sense 😅 

Does make me wonder though, should the related objects be in `syncStatus` rather than in the create? What's the benefit of having it there?

I'm concerned that calling update status multiple times might cause 409 conflict responses from the API when reconciling, have you seen that at all?

enxebre: >Does make me wonder though, should the related objects be in syncStatus rather than in the create? What's the benefit of having it there?
I'm concerned that calling update status multiple times might cause 409 conflict responses from the API when reconciling, have you seen that at all?

Make sense, let me do that.

clusteroperator.crd.yaml
```diff
@@ -0,0 +1,165 @@
...
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
```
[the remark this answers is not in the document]
enxebre: This is a snapshot just so the tests have the CRD available https://github.com/openshift/cluster-machine-approver/pull/71/commits/3ee0a6882cd4d81e5917b64d8c000c3982251bef#diff-390d58653a7a71d7b34e4ef2d39ae697R1

ObjectMeta is an api machinery top level known field for every resource.

status_test.go
JoelSpeed: Did you try this suggestion? I think it might work now the status has been refactored? WDYT?

clusteroperator.crd.yaml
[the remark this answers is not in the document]
elmiko: got it, thanks

i know ObjectMeta is all over the place, i was just kinda curious if these external schemas should use json pointers to reference back.

PR #72, opened 2020-05-21
status: Add AsExpected reasons

wking:
We've used empty-string reasons since the status handling landed in ed1c8070d5 (#71).  But if we feel like we have a message we want to set to help humans understand the available condition, we should be setting a reason string for machines too.  `AsExpected` follows [the library-go precedent][1].

Also use `AsExpected` for some message-less but expected conditions, like `Progressing=False` and `Degraded=False`.  Having a reason makes it easier for folks skimming logs to see that the condition is in its happy state, and doesn't cost us any effort to set.

I am looking forward to the smarter logic promised by 91a58e6619 (#71), because the current blind assumption that everything is working seems... very optimistic ;).

[1]: https://github.com/openshift/library-go/blob/94c59dec54be25c8527e51e8c0a885712aeb01b5/pkg/operator/status/condition.go#L67

### DIFF
--- a/status.go
+++ b/status.go
@@ -30,6 +30,7 @@ const (
 	unknownVersionValue           = "unknown"
 	queueKey                      = "trigger"
 	operatorVersionKey            = "operator"
+	reasonAsExpected              = "AsExpected"
 	releaseVersionEnvVariableName = "RELEASE_VERSION"
 )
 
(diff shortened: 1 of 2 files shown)


status.go
```diff
@@ -186,28 +186,28 @@ func (c *statusController) statusAvailable() error {
 			Type:               osconfigv1.OperatorAvailable,
 			Status:             osconfigv1.ConditionTrue,
 			LastTransitionTime: metav1.Now(),
-			Reason:             "",
+			Reason:             "AsExpected",
```
enxebre, line 189: thanks @wking!
can we please move this into a constant? Then in the test for each condition we do something like

```
Expect(v1helpers.FindStatusCondition(co.Status.Conditions, osconfigv1.OperatorAvailable).Reason).To(Equal(reasonAsExpected))```

enxebre:
>I am looking forward to the smarter logic promised by 91a58e6 (#71), because the current blind assumption that everything is working seems... very optimistic ;).

fwiw logic is already there and we short circuit the controller and alert on it. We now need to reuse that logic to signal on the operator status now. https://issues.redhat.com/browse/OCPCLOUD-876

status.go
wking: > ...can we please move this into a constant? Then in the tests for each condition...

Done with 9ec8de63 -> d86f36a6.

enxebre:
Thanks!
/retest
/approve
@elmiko @JoelSpeed PTAL for the ginkgo stanza

PR #75, opened 2020-06-15
add a metrics document

elmiko:
This change brings in a new document that describes the available
metrics.


elmiko:
> Are these developer docs or user docs?

imo, these are more developer facing documents.

PR #69, opened 2020-04-06
Drop glide in favour of go mod

enxebre:
This drops glide in favour of go mod: I run go mod init, go mod tidy, go mod vendor.
To get api machinery transitive dependencies happy I took the chance to drop deprecated openshift/cluster-api in favour of openshift/machine-api-operator.
As a consequence I also bumped to golang 1.13.

PR #76, opened 2020-07-07
[Release-4.4] BUG 1854385: Update golang/crypto to bring CVE-2020-9283

JoelSpeed:
This library is not used here for doing ssh client/server connections 
though still want to bring the library version with recent security 
fixes. https://github.com/golang/crypto/commit/bac4c82


enxebre:
have we considered to use this to move to go mod as in https://github.com/openshift/cluster-machine-approver/pull/69?

JoelSpeed:
If you think that's a better approach, I'm happy to do that, was hoping this would be a smaller diff but glide is not making this easy

JoelSpeed:
I cherry-picked the 4.5 switch to go.mod and then fought it into working with this codebase by adding a bunch of replacements. Seems to build on my machine now.

enxebre:
>I cherry-picked the 4.5 switch to go.mod 

this would bring kube 1.18 libraries here though, right?

>by adding a bunch of replacements

You mean in go mod?

JoelSpeed:
> this would bring kube 1.18 libraries here though, right?

It would, except after cherry picking, I set up go mod replacements to point all the kube dependencies back to the ones they were on before
https://github.com/openshift/cluster-machine-approver/pull/76/files#diff-37aff102a57d3d7b797f152915a6dc16R36-R46

PR #80, opened 2020-07-30
Document increased maxMachineDelta

pierreprinetti:
The maximum difference between the machine creation timestamp and the
CSR creation timestamp has been increased to accomodate for baremetal
machines. The README still reported "10 minutes" instead of "2 hours".

ref.: https://github.com/openshift/cluster-machine-approver/pull/37


elmiko:
thanks!
/approve

PR #83, opened 2020-09-02
Bug 1872080: Create Dockerfile.rhel file

enxebre:
This creates a Dockerfile.rhel file according to convention and symlink Dockerfile.rhel7 to the new one to satisfy tooling consumers until we switch them

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -1,0 +1,17 @@
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-openshift-4.6 AS builder
+WORKDIR /go/src/github.com/openshift/cluster-machine-approver
+COPY . .
+RUN make build
+
+FROM registry.svc.ci.openshift.org/ocp/4.6:base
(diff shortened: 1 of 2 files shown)


Danil-Grigorev:
/approve 
Nit: change commit message `sCreate Dockerfile.rhel file` -> `Create Dockerfile.rhel file`

elmiko:
thanks Alberto!
/lgtm

elmiko:
adding a reference to the dependent PR, https://github.com/openshift/ocp-build-data/pull/636

michaelgugino:
There's no need to do this.  We should update the release repo to point to docker.rhel once all the repos are migrated, then we remove the rhel7 dockerfile.

enxebre:
>There's no need to do this. We should update the release repo to point to docker.rhel once all the repos are migrated, then we remove the rhel7 dockerfile.

That needs to be done regardless. This just makes the current PR changes visible for any consumer tooling pointing to the deprecated file immediately both on CI against this PR and as they get merged afterwards.

PR #85, opened 2020-09-15
Bug 1879159: stop communicating with Kube API via localhost

p0lyn0mial:
Communicating with Kube API via localhost is discouraged because the caller doesn't know when the server is ready to serve the request. Reaching over the Kubernetes build-in service is preferred as it only contains the servers that are healthy and ready.

The following query shows that the operator frequently fails with errors like `127.0.0.1:6443: connect: connection refused` or `Failed to list *v1beta1.CertificateSigningRequest: certificatesigningrequests.certificates.k8s.io is forbidden: User "system:serviceaccount:openshift-cluster-machine-approver:machine-approver-sa" cannot list resource "certificatesigningrequests" in API group "certificates.k8s.io" at the cluster scope`

https://search.ci.openshift.org/?search=Failed+to+list+.*+is+forbidden%3A+.*+in+API+group&maxAge=48h&context=1&type=junit&name=&maxMatches=5&maxBytes=20971520&groupBy=job

from now on it should go via the kubernetes svc endpoint

### DIFF
--- a/manifests/04-deployment.yaml
+++ b/manifests/04-deployment.yaml
@@ -68,10 +68,6 @@ spec:
         env:
         - name: RELEASE_VERSION
           value: "0.0.1-snapshot"
-        - name: KUBERNETES_SERVICE_PORT
-          value: "6443"
-        - name: KUBERNETES_SERVICE_HOST
-          value: "127.0.0.1"
         - name: METRICS_PORT
           value: "9191"
         terminationMessagePolicy: FallbackToLogsOnError


manifests/04-deployment.yaml
```diff
@@ -20,7 +20,6 @@ spec:
       labels:
         app: machine-approver
     spec:
-      hostNetwork: true
```
sttts, line 23: this is probably fine, at least from our point of view.

sttts: otherwise, this will depend on SDN networking which might be undesired.

enxebre:
/retitle Bug 1879159: stop communicating with Kube API via localhost

manifests/04-deployment.yaml
Danil-Grigorev: This is the only place where CSR approval flow will experience any changes. We don't use `KUBERNETES_SERVICE_HOST` explicitly anywhere, so as `127.0.0.1`. I wonder if there is a way to check the health of the API server before starting `CSR` serving.

p0lyn0mial: LBs watch `/readyz` endpoint as described in https://github.com/openshift/installer/blob/master/docs/dev/kube-apiserver-health-check.md so maybe this is something the operator could rely on as well.

Wondering what would this operator do when the local API server rolls out to a new version?

sttts: why 127.0.0.1 in the first place? This feels wrong. Don't use it.

enxebre: @p0lyn0mial As it's late in the release cycle can we keep hostNetwork to reduce the scope of the change and  move forward with the PR?

enxebre: @p0lyn0mial friendly ping :) so we can close anything open for 4.6

p0lyn0mial: yes, sorry, updating the PR.

p0lyn0mial: done :)

PR #73, opened 2020-06-08
[OCPCLOUD-893] Refactor CSR handling logic to utilize controller-runtime

Danil-Grigorev:
https://issues.redhat.com/browse/OCPCLOUD-893

PR #95, opened 2020-11-19
[OCPCLOUD-1042] Increase test coverage in CSR approval

Danil-Grigorev:
Porting introduced testst from https://github.com/openshift/cluster-machine-approver/pull/73/commits/a7a33f99fe82374879b54d9fa558628fd9e9ff5d as a part of https://github.com/openshift/cluster-machine-approver/pull/73, which are adding full coverage of `csr_check.go` package.
```
Before
ok  	github.com/openshift/cluster-machine-approver	7.143s	coverage: 46.7% of statements
After
ok  	github.com/openshift/cluster-machine-approver	(cached)	coverage: 63.7% of statements
```

Depends on https://github.com/openshift/cluster-machine-approver/pull/96

### DIFF
--- a/csr_check.go
+++ b/csr_check.go
@@ -20,7 +20,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 )
 
@@ -44,7 +43,13 @@ var nodeBootstrapperGroups = sets.NewString(
 	"system:authenticated",
 )
 
+var now = time.Now
+
 var maxPendingCSRs uint32
+
+type csrLister interface {
+	List() []interface{}
+}
 
 func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr *x509.CertificateRequest) (string, error) {
 	if !strings.HasPrefix(req.Spec.Username, nodeUserPrefix) {
@@ -75,14 +80,9 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
 		return "", fmt.Errorf("Too few usages")
 	}
 
-	usages := make([]string, 0)
+	usages := make([]string, 3)
 	for i := range req.Spec.Usages {
-		usages = append(usages, string(req.Spec.Usages[i]))
-	}
-
-	// No extra usages!
-	if len(usages) != 3 {
-		return "", fmt.Errorf("Unexpected usages: %d", len(usages))
+		usages[i] = string(req.Spec.Usages[i])
 	}
 
 	usageSet := sets.NewString(usages...)
@@ -162,7 +162,7 @@ func authorizeCSR(
 		if err == nil && servingCert != nil {
 			klog.Infof("Found existing serving cert for %s", nodeAsking)
 
-			err := authorizeServingRenewal(nodeAsking, csr, servingCert, ca)
+			err := authorizeServingRenewal(nodeAsking, csr, servingCert, x509.VerifyOptions{Roots: ca})
 
 			// No error, the renewal is authorized.
 			if err == nil {
@@ -290,14 +290,15 @@ func authorizeNodeClientCSR(config ClusterMachineApproverConfig, machines []v1be
 // The current certificate must be signed by the current CA and not expired.
 // The common name on the current certificate must match the expected value.
 // All Subject Alternate Name values must match between CSR and current cert.
-func authorizeServingRenewal(nodeName string, csr *x509.CertificateRequest, currentCert *x509.Certificate, ca *x509.CertPool) error {
-	if csr == nil || currentCert == nil || ca == nil {
+func authorizeServingRenewal(nodeName string, csr *x509.CertificateRequest, currentCert *x509.Certificate, options x509.VerifyOptions) error {
+	// options.Roots should contain root certificates
+	if csr == nil || currentCert == nil || options.Roots == nil {
 		return fmt.Errorf("CSR, serving cert, or CA not provided")
 	}
 
 	// Check that the serving cert is signed by the given CA, is not expired,
 	// and is otherwise valid.
-	if _, err := currentCert.Verify(x509.VerifyOptions{Roots: ca}); err != nil {
+	if _, err := currentCert.Verify(options); err != nil {
 		return err
 	}
 
@@ -361,15 +362,15 @@ func isApproved(csr *certificatesv1beta1.CertificateSigningRequest) bool {
 	return false
 }
 
-func recentlyPendingCSRs(indexer cache.Indexer) int {
+func recentlyPendingCSRs(lister csrLister) int {
 	// assumes we are scheduled on the master meaning our clock is the same
-	now := time.Now()
-	start := now.Add(-maxPendingDelta)
-	end := now.Add(maxMachineClockSkew)
+	currentTime := now()
+	start := currentTime.Add(-maxPendingDelta)
+	end := currentTime.Add(maxMachineClockSkew)
 
 	var pending int
 
-	for _, item := range indexer.List() {
+	for _, item := range lister.List() {
 		csr := item.(*certificatesv1beta1.CertificateSigningRequest)
 
 		// ignore "old" CSRs
--- a/csr_check_test.go
+++ b/csr_check_test.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"crypto/x509"
-	stderrors "errors"
+	"encoding/pem"
+	"fmt"
 	"net"
 	"net/url"
 	"reflect"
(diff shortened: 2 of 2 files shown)


Danil-Grigorev:
/hold 

This PR requires golang-1.15 in CI. Otherwise we get an error message mismatch:
Expected
```x509: certificate has expired or is not yet valid: current time 2020-11-18T00:00:00Z is before 2020-11-18T20:12:00Z```
and CI returns just
```x509: certificate has expired or is not yet valid```

File: csr_check_test.go [hunk 1]
```diff
@@ -1220,8 +1402,7 @@ func Test_authorizeCSR(t *testing.T) {
 						},
 					},
 				},
-				nodeName: "bear",
-				nodeErr:  errors.NewNotFound(schema.GroupResource{}, ""),
+				node: withName("bear", defaultNode()),
```
elmiko, line 1405: i'm curious why some of these entries have the `nodeErr` removed?

csr_check_test.go
```diff
@@ -1992,6 +2355,177 @@ func TestAuthorizeServingRenewal(t *testing.T) {
...
+			name:      "wrong address",
+			nodeName:  "test",
+			node:      wrongAddr,
+			rootCerts: []*x509.Certificate{parseCert(t, rootCertGood)},
+			wantErr:   "dial tcp 127.0.0.1:25544: connect: connection refused",
+		},
+		{
+			name:     "no pool provided",
+			nodeName: "test",
+			node:     wrongAddr,
```
elmiko, line 2421: should this be `wrongAddr`? 

i might be slightly confused byt the test name/logic

csr_check_test.go
```diff
@@ -1992,6 +2355,177 @@ func TestAuthorizeServingRenewal(t *testing.T) {
...
+			if len(tt.rootCerts) > 0 {
+				certPool = x509.NewCertPool()
+				for _, cert := range tt.rootCerts {
+					certPool.AddCert(cert)
+				}
+			}
+
+			nodes := &testNode{t: t, node: tt.node, err: tt.serverErr}
+
+			go respond(server)
```
elmiko, line 2448: is there any chance we would deadlock here inside of `respond`? (i saw an `accept` call in that function)

csr_check_test.go
```diff
@@ -1940,49 +2111,241 @@ func Test_authorizeCSR(t *testing.T) {
...
 
-			if err := authorizeCSR(tt.args.config, tt.args.machines, nodes, tt.args.req, parsedCSR, tt.args.ca); errString(err) != tt.wantErr {
+			var ca *x509.CertPool
+			if len(tt.args.ca) > 0 {
+				// Start renewal flow
+				ca = x509.NewCertPool()
+				for _, cert := range tt.args.ca {
+					ca.AddCert(cert)
+				}
+				go respond(kubeletServer)
```
elmiko, line 2250: is there any chance we would deadlock here inside of `respond`? (i saw an `accept` call in that function)

elmiko (commented):
this is _mostly_ making sense to me Danil, i just have a few questions to make it clearer for me.

csr_check_test.go
Danil-Grigorev: Accept returns an error, once the socket is closed. I ignore it, as I only needed a consumer to return some response with peer certificates (if it is needed, and I call it only when it is needed)

csr_check_test.go
Danil-Grigorev: https://github.com/openshift/cluster-machine-approver/pull/95#discussion_r527597436

csr_check.go
```diff
@@ -361,23 +357,21 @@ func isApproved(csr *certificatesv1beta1.CertificateSigningRequest) bool {
 	return false
 }
 
-func recentlyPendingCSRs(indexer cache.Indexer) int {
+func recentlyPendingCSRs(csrs []certificatesv1beta1.CertificateSigningRequest) int {
 	// assumes we are scheduled on the master meaning our clock is the same
-	now := time.Now()
+	now := now()
```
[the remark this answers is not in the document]
Danil-Grigorev: Right, we need to test the certificate exchange flow, and so I had to stub `time.Now` -https://github.com/openshift/cluster-machine-approver/pull/95/files#diff-3c8388bac66caa11eeb9a56a4cb51f879caff503f20824b7bd973f7969eda6ddR412

csr_check_test.go
Danil-Grigorev: These tests need a lot of refactoring. These test cases as you can see contain no changes aside from this, and they don't fail. So it was initially a copy-paste job, and it is a dead code.

csr_check.go
```diff
@@ -75,14 +76,9 @@ func validateCSRContents(req *certificatesv1beta1.CertificateSigningRequest, csr
 		return "", fmt.Errorf("Too few usages")
```
JoelSpeed, line 76: This error message could really do with improvement couldn't it 🤔

csr_check.go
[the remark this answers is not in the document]
JoelSpeed: Can we use a name that doesn't shadow the top level please?

main.go
```diff
@@ -157,7 +157,11 @@ func (c *Controller) handleNewCSR(key string) error {
 	}
 	maxPending := getMaxPending(machines.Items)
 	atomic.StoreUint32(&maxPendingCSRs, uint32(maxPending))
-	if pending := recentlyPendingCSRs(c.indexer); pending > maxPending {
+	csrs := []*certificatesv1beta1.CertificateSigningRequest{}
+	for _, item := range c.indexer.List() {
+		csrs = append(csrs, item.(*certificatesv1beta1.CertificateSigningRequest))
+	}
+	if pending := recentlyPendingCSRs(csrs); pending > maxPending {
```
JoelSpeed, line 164: What's the motivation for moving this outside of the function? I'm not a huge fan of moving logic around to suit the test suites, there are better ways we can do this, eg

```
type indexLister interface {
  List() []interface{}
}

func recentlyPendingCSRs(csrs indexLister) []*CertificateSigningRequest {
...
}

// In test code

type fakeIndexLister struct {
 items []interface{}
}

func (f fakeIndexList) List() ([]interface{}) {
  return f.items
}
```

metrics.go
```diff
@@ -39,7 +40,11 @@ func (mc MetricsCollector) Describe(ch chan<- *prometheus.Desc) {
 
 // Collect implements the prometheus.Collector interface.
 func (mc MetricsCollector) collectMetrics(ch chan<- prometheus.Metric) {
-	pending := recentlyPendingCSRs(mc.cacheIndexer)
+	csrs := []*certificatesv1beta1.CertificateSigningRequest{}
+	for _, item := range mc.cacheIndexer.List() {
+		csrs = append(csrs, item.(*certificatesv1beta1.CertificateSigningRequest))
+	}
```
JoelSpeed, line 46: The fact we had to move this twice, definitely suggests we should be making the tests bend to the code rather than the other way around

csr_check_test.go
```diff
@@ -1992,6 +2355,177 @@ func TestAuthorizeServingRenewal(t *testing.T) {
...
+				createdAt(pastApprovalTime, pendingCSR),
+			},
+			expectPending: 2,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if pending := recentlyPendingCSRs(tt.csrs); pending != tt.expectPending {
+				t.Errorf("Expected %v pending CSRs, got: %v", tt.expectPending, pending)
```
JoelSpeed, line 2523: Looking at the way these tests are structured, we could easily use a fake lister here to avoid having to move the logic in the main code flow

csr_check_test.go
```diff
@@ -2296,3 +2873,48 @@ func assertNoChange(t *testing.T, a, b []string, f func(*testing.T)) {
...
+		GetCertificate: func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) { return &crt, nil },
+	})
+	if err != nil {
+		t.Fatalf("Fail to establish TCP listener: %s", err.Error())
+	}
+
+	return
+}
+
+func respond(server net.Listener) {
```
JoelSpeed, line 2911: Am I right in thinking `Accept` will return an error if the listener is shutdown?

main.go
Danil-Grigorev: The goal of this PR is to transition to https://github.com/openshift/cluster-machine-approver/pull/73 and do not change existing tests. The thing you pointing out makes sense to me, but I didn't think about it as a long-term solution, just wanted to preserve tests.

JoelSpeed: I don't think it would take too long to do, and would be the right thing given how long it might take to merge things across to that other PR, but it's not a blocker if that's going away in #73

csr_check_test.go
Danil-Grigorev: It will. I just needed a consumer, so ignoring all the errors worked for me in such a scenario. One pass is guaranteed to be a success, so the code accepts and responds only when it's needed.

csr_check.go
Danil-Grigorev: I'll work on this in https://github.com/openshift/cluster-machine-approver/pull/73

elmiko (approved):
i'm good with this, the only outstanding question i have is did Joel's request for changing the name on `now` get picked up?
/lgtm

JoelSpeed:
/approve

Please make sure that #73 picks up the outstanding changes that I had asked for re not having to move logic out of functions to help tests

PR #101, opened 2021-01-05
[OCPCLOUD-893] Refactor csr approver to use controller runtime

Danil-Grigorev:
Moving out old controller parts of csr approver, and replacing them with the controller-runtime.


michaelgugino:
This is failing to run:
```
cluster-machine-approver flag redefined: log_dir
panic: cluster-machine-approver flag redefined: log_dir
```

pkg/metrics/metrics.go
```diff
@@ -39,8 +39,7 @@ func (mc MetricsCollector) Describe(ch chan<- *prometheus.Desc) {
 
 // Collect implements the prometheus.Collector interface.
 func (mc MetricsCollector) collectMetrics(ch chan<- prometheus.Metric) {
-	pending := recentlyPendingCSRs(mc.cacheIndexer)
-	ch <- prometheus.MustNewConstMetric(CurrentPendingCSRCountDesc, prometheus.GaugeValue, float64(pending))
-	ch <- prometheus.MustNewConstMetric(MaxPendingCSRDesc, prometheus.GaugeValue, float64(atomic.LoadUint32(&maxPendingCSRs)))
+	ch <- prometheus.MustNewConstMetric(CurrentPendingCSRCountDesc, prometheus.GaugeValue, float64(atomic.LoadUint32(&controller.PendingCSRs)))
+	ch <- prometheus.MustNewConstMetric(MaxPendingCSRDesc, prometheus.GaugeValue, float64(atomic.LoadUint32(&controller.MaxPendingCSRs)))
```
JoelSpeed, line 43: We really should remove this custom metrics collection logic and just use prometheus gauges and observe the values inline

JoelSpeed (commented):
/hold This will need to wait until master opens for 4.8 before we can merge. Once master opens let's make sure it merges quick so it has plenty of time to soak.

Found a couple of issues while reading through that will need to be fixed, I might suggest we look at creating a follow up PR for changing the metrics on this, WDYT? 

Equally, would be good to get a follow up to upgrade controller-runtime to v0.7.0, shouldn't be too complicated but we will want to pass the new context through the reconcile methods so we don't randomly use `context.Background()` everywhere (Nit: should really be using `context.TODO()` in controller code to make sure we eventually use a real context)

pkg/controller/controller.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+	}
+
+	return reconcile.Result{}, fmt.Errorf("Failed to find CSR: %v", req)
+}
+
+// reconcileLimits will short circut logic if number of pending CSRs is exceeding limit
+func reconcileLimits(machines *machinev1.MachineList, csrs *certificatesv1.CertificateSigningRequestList) bool {
+	maxPending := getMaxPending(machines.Items)
+	atomic.StoreUint32(&MaxPendingCSRs, uint32(maxPending))
```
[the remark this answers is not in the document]
Danil-Grigorev: SGTM, but I think it is out of scope for this PR. Will be glad to pick it up in a followup.

Danil-Grigorev:
@JoelSpeed PTAL - feedback is adressed.

Makefile
```diff
@@ -14,7 +14,7 @@ test:
 .PHONY: test
 
 unit:
-	go test -v .
+	go test -v -p 1 ./...
```
JoelSpeed, line 17: Do we need to specify the `-p` flag here? We don't do it normally

pkg/controller/controller.go
```diff
@@ -0,0 +1,238 @@
...
+			GenericFunc: func(e event.GenericEvent) bool { return pendingCertFilter(e.Object) },
+			DeleteFunc:  func(e event.DeleteEvent) bool { return false },
+		})).
+		Watches(
+			&source.Kind{Type: &corev1.ConfigMap{}},
+			handler.EnqueueRequestsFromMapFunc(m.toCSRs),
+			builder.WithPredicates(predicate.Funcs{
+				CreateFunc:  func(e event.CreateEvent) bool { return caConfigMapFilter(e.Object, nil) },
+				UpdateFunc:  func(e event.UpdateEvent) bool { return caConfigMapFilter(e.ObjectOld, e.ObjectNew) },
+				GenericFunc: func(e event.GenericEvent) bool { return pendingCertFilter(e.Object) },
```
JoelSpeed, line 62: Should this one be `caConfigMapFilter` too? Potential copy/paste error

Makefile
Danil-Grigorev: I noticed occasional unit flakes, which were part of previous test code, due to attempts to operate on a single ClusterOperator resource in parallel, and not allowing it to clean results from previous execution.

JoelSpeed: Is the default not to run tests serially anyway? I had thought this change isn't actually doing anything when I first read it

pkg/controller/controller.go
Danil-Grigorev: Fixed, thanks @JoelSpeed

JoelSpeed:
Please make sure to squash your commits down to a set of logical atomic commits before we merge, quite a few fixup style commits on the end of this at the moment

Makefile
JoelSpeed: Still not sure why this makes a difference, have we declared explicitly that some of our tests are parallel?

Danil-Grigorev: You are right, they are. I just checked. We will probably just need something which ensures the resource was actually deleted. Depends on the frequency of the flakes.

JoelSpeed: Let's remove this for now and fix in a separate PR

JoelSpeed:
/hold cancel

Let's check the tests all pass, but LGTM initially

PR #108, opened 2021-03-19
Refactor logging and error handling

michaelgugino:
Currently, logging is almost nonexistent.  Returning
errors to the controller manager does not result in viewable
messages on current verbosity.  This commit adds explicit
klog statements for better debugging and tracability.

Currently, we return error to the controller manager too often.  This
results in CSRs being endlessly requeued which will never be approved.
This commit less-eagerly requeues CSRs for most permanent failure modes.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -111,16 +111,19 @@ func caConfigMapFilter(obj runtime.Object, new runtime.Object) bool {
 
 func (m *CertificateApprover) Reconcile(ctx context.Context, req ctrl.Request) (reconcile.Result, error) {
 	csrs := &certificatesv1.CertificateSigningRequestList{}
+	klog.Infof("Reconciling CSR: %v", req.Name)
 	if err := m.List(ctx, csrs); err != nil {
+		klog.Errorf("%v: Failed to list CSRs: %v", req.Name, err)
 		return reconcile.Result{}, fmt.Errorf("Failed to get CSRs: %w", err)
 	}
 
 	machines := &machinev1.MachineList{}
 	if err := m.List(ctx, machines); err != nil {
+		klog.Errorf("%v: Failed to list machines: %v", req.Name, err)
 		return reconcile.Result{}, fmt.Errorf("Failed to list machines: %w", err)
 	}
 
-	if offLimits := reconcileLimits(machines, csrs); offLimits {
+	if offLimits := reconcileLimits(req.Name, machines, csrs); offLimits {
 		// Stop all reconciliation
 		return reconcile.Result{}, nil
 	}
@@ -137,13 +140,13 @@ func (m *CertificateApprover) Reconcile(ctx context.Context, req ctrl.Request) (
 }
 
 // reconcileLimits will short circut logic if number of pending CSRs is exceeding limit
-func reconcileLimits(machines *machinev1.MachineList, csrs *certificatesv1.CertificateSigningRequestList) bool {
+func reconcileLimits(csrName string, machines *machinev1.MachineList, csrs *certificatesv1.CertificateSigningRequestList) bool {
 	maxPending := getMaxPending(machines.Items)
 	atomic.StoreUint32(&MaxPendingCSRs, uint32(maxPending))
 	pending := recentlyPendingCSRs(csrs.Items)
 	atomic.StoreUint32(&PendingCSRs, uint32(pending))
 	if pending > maxPending {
-		klog.Errorf("Pending CSRs: %d; Max pending allowed: %d. Difference between pending CSRs and machines > %v. Ignoring all CSRs as too many recent pending CSRs seen", pending, maxPending, maxDiffBetweenPendingCSRsAndMachinesCount)
+		klog.Errorf("%v: Pending CSRs: %d; Max pending allowed: %d. Difference between pending CSRs and machines > %v. Ignoring all CSRs as too many recent pending CSRs seen", csrName, pending, maxPending, maxDiffBetweenPendingCSRsAndMachinesCount)
 		return true
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -163,9 +167,10 @@ func (m *CertificateApprover) reconcileCSR(csr certificatesv1.CertificateSigning
 		klog.Errorf("failed to get kubelet CA")
 	}
 
-	if err := authorizeCSR(m, m.Config, machines.Items, &csr, parsedCSR, kubeletCA); err != nil {
+	if authorize, err := authorizeCSR(m, m.Config, machines.Items, &csr, parsedCSR, kubeletCA); !authorize {
 		// Don't deny since it might be someone else's CSR
-		return fmt.Errorf("CSR %s not authorized: %v", csr.Name, err)
+		klog.Infof("%s: CSR not authorized", csr.Name)
+		return err
```
JoelSpeed, line 173: Will the `err` not contain any useful information?

pkg/controller/csr_check.go
```diff
@@ -131,20 +133,30 @@ func authorizeCSR(
...
 	if req == nil || csr == nil {
-		return fmt.Errorf("Invalid request")
+		klog.Errorf("authorizeCSR invalid request")
+		return false, nil
 	}
 
 	if isNodeClientCert(req, csr) {
-		return authorizeNodeClientCSR(c, config, machines, req, csr)
+		if config.NodeClientCert.Disabled {
+			klog.Errorf("%v: CSR rejected as the flow is disabled", req.Name)
```
JoelSpeed, line 144: Do we want to point out here we have identified it as a node client cert? What are the security implications of logging that?

pkg/controller/csr_check.go
```diff
@@ -131,20 +133,30 @@ func authorizeCSR(
...
+		return authorizeNodeClientCSR(c, machines, req, csr)
 	}
 
+	klog.Infof("%v: CSR does not appear to be client csr", req.Name)
 	// node serving cert validation after this point
 
 	nodeAsking, err := validateCSRContents(req, csr)
-	if err != nil {
-		return err
+	if nodeAsking == "" {
```
JoelSpeed, line 154: What happens to the `err` here? Why are we ignoring it? Will it always be nil if `nodeAsking` is populated? Can that be guaranteed into the future? Maybe check it explicitly as well?
```suggestion
	if nodeAsking == "" || err != nil {
```

pkg/controller/csr_check.go
```diff
@@ -207,7 +221,11 @@ func authorizeCSR(
 		}
 		// The CSR requested a DNS name that did not belong to the machine
 		if !foundSan {
-			return fmt.Errorf("DNS name '%s' not in machine names: %s", san, strings.Join(attemptedAddresses, " "))
+			//TODO: set annotation/emit event here.
+			// return error so we requeue, in case machine network is out of date
+			// for some reason
+			klog.Errorf("%v: DNS name '%s' not in machine names: %s", req.Name, san, strings.Join(attemptedAddresses, " "))
```
JoelSpeed, line 227: Are there security implications of logging things such as DNS names?

pkg/controller/csr_check.go
```diff
@@ -231,47 +249,63 @@ func authorizeCSR(
 		}
 		// The CSR requested an IP name that did not belong to the machine
 		if !foundSan {
-			return fmt.Errorf("IP address '%s' not in machine addresses: %s", san, strings.Join(attemptedAddresses, " "))
+			//TODO: set annotation/emit event here.
+			// return error so we requeue, in case machine network is out of date
+			// for some reason
+			klog.Errorf("%v: IP address '%s' not in machine addresses: %s", req.Name, san, strings.Join(attemptedAddresses, " "))
```
JoelSpeed, line 255: Are there security implications of logging things like IP addresses?

pkg/controller/csr_check.go
```diff
@@ -231,47 +249,63 @@ func authorizeCSR(
...
+		klog.Errorf("%v: machine for node %s already has node ref, cannot approve", req.Name, nodeName)
+		return false, nil
 	}
 
 	start := nodeMachine.CreationTimestamp.Add(-maxMachineClockSkew)
 	end := nodeMachine.CreationTimestamp.Add(maxMachineDelta)
 	if !inTimeSpan(start, end, req.CreationTimestamp.Time) {
-		return fmt.Errorf("CSR %s creation time %s not in range (%s, %s)", req.Name, req.CreationTimestamp.Time, start, end)
+		//TODO: set annotation/emit event here.
+		klog.Errorf("%v: CSR creation time %s not in range (%s, %s)", req.Name, req.CreationTimestamp.Time, start, end)
```
JoelSpeed, line 304: Could putting this timing information in here be used later to forge a valid CSR if these logs were somehow leaked?

JoelSpeed (commented):
In general LGTM, just a few questions about the particular details we are logging. I want us to be sure that by logging details from the CSRs we aren't opening ourselves to future attacks if the logs were somehow accessed by an attacker trying to forge a valid CSR

pkg/controller/csr_check.go
michaelgugino: No.  DNS names appear in non-secret objects all over the place.  The public cert is just that, public.

pkg/controller/csr_check.go
michaelgugino: No.

pkg/controller/csr_check.go
michaelgugino: No.

pkg/controller/controller.go
michaelgugino: The error will have already been logged inside the function.

pkg/controller/csr_check.go
michaelgugino: There are no security implications.

PR #111, opened 2021-04-12
Bug 1947785: Move certificates to v1

alexander-demichev:
Move certificates to v1 because v1beta1 will be removed in 4.9.


elmiko (approved):
/lgtm

JoelSpeed:
/hold This isn't working correctly, if you look at the CSRs, they aren't being approved, even though the CSR approver is reconciling them. Something in this change breaks the approving/updating logic but there isn't logging to tell us what it might be.

This will need further investigation

PR #50, opened 2019-09-26
Metrics support

spangenberg:
Taking over https://github.com/openshift/cluster-machine-approver/pull/34 in order to solve merge conflicts.

PR #112, opened 2021-04-13
manifests/0000_90_cluster-machine-approver_04_alertrules: Drop ClusterMachineApproverDown

wking:
The alert was added in 2e7f847b29 (#50).  But the cluster-version operator is responsible for complaining if the machine approver operator's deployment is sad, so no need for the operator to handle this directly (we end up doubling up if there's an issue).

Similar to openshift/cloud-credential-operator#308, openshift/cluster-autoscaler-operator#196, and openshift/machine-api-operator#826.


elmiko (approved):
/approve

PR #114, opened 2021-04-15
Add support for unstructured machines

alexander-demichev:
In order to make cluster machine approver work with CAPI and MAPI machines, I converted the code to use unstructured. The API group is passed as a CLI argument and defaults to MAPI. The unstructured approach is similar to what is done in autoscaler.

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,32 +21,44 @@ import (
 	"fmt"
 	"os"
 	"strconv"
-	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-machine-approver/pkg/controller"
 	"github.com/openshift/cluster-machine-approver/pkg/metrics"
-	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
 	control "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
 
-	"k8s.io/klog/v2"
+const (
+	capiGroup = "cluster.x-k8s.io"
+	mapiGroup = "machine.openshift.io"
 )
 
 func main() {
 	var cliConfig string
 	var APIGroup string
+	var managementKubeConfigPath string
+	var workloadKubeConfigPath string
 
 	flagSet := flag.NewFlagSet("cluster-machine-approver", flag.ExitOnError)
 
 	klog.InitFlags(flagSet)
 	flagSet.StringVar(&cliConfig, "config", "", "CLI config")
+	flagSet.StringVar(&APIGroup, "apigroup", "machine.openshift.io", "API group for machines, defaults to machine.openshift.io")
+	flagSet.StringVar(&managementKubeConfigPath, "management-cluster-kubeconfig", "", "management kubeconfig path,")
+	flagSet.StringVar(&workloadKubeConfigPath, "workload-cluster-kubeconfig", "", "workload kubeconfig path")
+
 	flagSet.Parse(os.Args[1:])
 
-	flagSet.StringVar(&APIGroup, "apigroup", "machine.openshift.io", "API group for machines, defaults to machine.openshift.io")
+	if err := validateAPIGroup(APIGroup); err != nil {
+		klog.Fatalf(err.Error())
+	}
 
 	// Now let's start the controller
 	stop := make(chan struct{})
@@ -61,12 +73,20 @@ func main() {
 		metricsPort = fmt.Sprintf(":%d", v)
 	}
 
+	managementConfig, workloadConfig, err := createClientConfigs(managementKubeConfigPath, workloadKubeConfigPath)
+	if err != nil {
+		klog.Fatalf("Can't set client configs: %v", err)
+	}
+
+	managementClient, workloadClient, err := createClients(managementConfig, workloadConfig)
+	if err != nil {
+		klog.Fatalf("Can't create clients: %v", err)
+	}
+
 	// Create a new Cmd to provide shared dependencies and start components
 	klog.Info("setting up manager")
-	syncPeriod := 10 * time.Minute
-	mgr, err := manager.New(control.GetConfigOrDie(), manager.Options{
+	mgr, err := manager.New(managementConfig, manager.Options{
 		MetricsBindAddress: metricsPort,
-		SyncPeriod:         &syncPeriod,
 	})
 	if err != nil {
 		klog.Fatalf("unable to set up overall controller manager: %v", err)
@@ -79,27 +99,30 @@ func main() {
 		klog.Fatal(err)
 	}
 
-	if err := machinev1.AddToScheme(mgr.GetScheme()); err != nil {
-		klog.Fatal("unable to add Machines to scheme")
-	}
-
-	directClient, err := client.New(mgr.GetConfig(), client.Options{
-		Scheme: mgr.GetScheme(),
-		Mapper: mgr.GetRESTMapper(),
-	})
-	if err != nil {
-		klog.Fatal("unable to set up client")
-	}
 	// Prevent the controller from caching node and machine objects.
 	// Stale nodes and machines can cause the approver to not approve certificates
 	// within a timely manner, leading to failed node bootstraps.
-	approverClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
-		Client:      directClient,
+	uncachedManagementClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
+		Client:      *managementClient,
 		CacheReader: mgr.GetClient(),
 		UncachedObjects: []client.Object{
-			&machinev1.Machine{},
 			&corev1.Node{},
 		},
+		// CacheUnstructured should be false because we manipulate with unstructured machines
+		CacheUnstructured: false,
+	})
+	if err != nil {
+		klog.Fatalf("unable to set up delegating client: %v", err)
+	}
+
+	uncachedWorkloadClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
+		Client:      *workloadClient,
+		CacheReader: mgr.GetClient(),
+		UncachedObjects: []client.Object{
+			&corev1.Node{},
+		},
+		// CacheUnstructured should be false because we manipulate with unstructured machines
+		CacheUnstructured: false,
 	})
 	if err != nil {
 		klog.Fatalf("unable to set up delegating client: %v", err)
@@ -108,9 +131,12 @@ func main() {
 	// Setup all Controllers
 	klog.Info("setting up controllers")
 	if err = (&controller.CertificateApprover{
-		Client:  approverClient,
-		RestCfg: mgr.GetConfig(),
-		Config:  controller.LoadConfig(cliConfig),
+		MachineClient:  uncachedManagementClient,
+		MachineRestCfg: managementConfig,
+		NodeClient:     uncachedWorkloadClient,
+		NodeRestCfg:    workloadConfig,
+		Config:         controller.LoadConfig(cliConfig),
+		APIGroup:       APIGroup,
 	}).SetupWithManager(mgr, ctrl.Options{}); err != nil {
 		klog.Fatalf("unable to create CSR controller: %v", err)
 	}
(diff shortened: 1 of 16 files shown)


pkg/machinehandler/machinehandler.go
```diff
@@ -0,0 +1,174 @@
...
+		Version: APIVersion,
+	})
+	if err := m.Client.List(m.Ctx, unstructuredMachineList); err != nil {
+		return err
+	}
+
+	m.machines = unstructuredMachineList.Items
+
+	return nil
+}
```
JoelSpeed, line 62: I wonder instead if we should create a "client.List" implementation that returns our own internal machine structure as a list

If we were to define a struct something like
```
type Machine struct {
  ObjectMeta metav1.ObjectMeta
  Status struct {
    NodeRef metav1.ObjectReference
    NodeAddresses corev1.NodeAddress
  }
}
```

And then used something like https://github.com/mitchellh/mapstructure, we could convert the unstructured data into our new type (preserving only the values that we care about) and the rest of the logic in the approver would be pretty simplistic and would just have to read from the fields as it does today

main.go
```diff
@@ -35,13 +35,16 @@ import (
...
 	var cliConfig string
+	var APIGroup string
 
 	flagSet := flag.NewFlagSet("cluster-machine-approver", flag.ExitOnError)
 
 	klog.InitFlags(flagSet)
 	flagSet.StringVar(&cliConfig, "config", "", "CLI config")
 	flagSet.Parse(os.Args[1:])
 
+	flagSet.StringVar(&APIGroup, "apigroup", "machine.openshift.io", "API group for machines, defaults to machine.openshift.io")
```
JoelSpeed, line 46: Do we want to include the kubeconfig versions changes that would be needed for split clusters in this PR or is that a follow up?

alexander-demichev: I'm working on it right now, can be later merged in a follow-up

pkg/machinehandler/machinehandler.go
alexander-demichev: that's a good idea, it can make the implementation cleaner.

main.go
```diff
@@ -95,3 +109,35 @@ func main() {
...
+	if managementClientPath == "" {
+		return control.GetConfigOrDie(), control.GetConfigOrDie(), nil
+	}
+
+	managementConfig, err := clientcmd.BuildConfigFromFlags("", managementClientPath)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return managementConfig, control.GetConfigOrDie(), nil
```
enxebre, line 126: would this get the workload kubeconfig in-cluster? This binary is expected to run in the management cluster so the workload config would come from a secret and need to be used to manipulate nodes and watch CSR.

JoelSpeed: IIUC we already have a `kubeconfig` flag which allows you to override the workload config, so I think that answers running in the management cluster with a secret pointing to the workload cluster, but then it seems we have to also specify the management cluster as a kubeconfig as well? 

I think we need to have some way for the management config to be used as in cluster creds, while the workload is using a secret based kubeconfig. I think in the autoscaler there's a special flag for this, we should copy their pattern

alexander-demichev: I tried to follow their pattern https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_provider.go#L145

In the case of default openshift deployment, nothing changes because management config == workload config. For the case when we need separate clients: management cluster config is obtained using `--management-kubeconfig` flag and workload using `--kubeconfig`.

main.go
```diff
@@ -125,3 +141,38 @@ func main() {
 		klog.Fatalf("unable to run the manager: %v", err)
 	}
 }
+
+// createClientConfigs allow users to provide second config using management-kubeconfig, if specified
+// try to build it from provided path. First returned value is management config used for Machines,
+// second is workload config used for Node/CSRs.
+func createClientConfigs(managementClientPath string, managementConfigAuthoritative bool) (*rest.Config, *rest.Config, error) {
+	managementConfig := control.GetConfigOrDie()
```
enxebre, line 149: If I'm reading this correctly the workloadConfig is always in-cluster?
We need the workloadConfig to be configurable. We want to run the machine approver in the management cluster

Also I know @JoelSpeed suggested the `managementConfigAuthoritative` pattern but I can't recall why is that necessary?

pkg/controller/controller.go
```diff
@@ -33,8 +33,12 @@ const (
 
 // MachineApproverReconciler reconciles a machine-approver  object
 type CertificateApprover struct {
-	client.Client
-	RestCfg  *rest.Config
+	NodeClient  client.Client
+	NodeRestCfg *rest.Config
```
enxebre, line 37: I know the NodeRestCfg is used by the `approve()` signature but seems unnecessary we we already have the clients here, so we can reuse the nodeClient or build a csrClient.

pkg/controller/controller.go
```diff
@@ -33,8 +33,12 @@ const (
...
+
+	MachineClient  client.Client
+	MachineRestCfg *rest.Config
```
enxebre, line 40: why do we need `MachineRestCfg`? we already have the client

main.go
enxebre: oh I see this would use --kubeconfig if set.
May be is only me but this pattern results very confusing.
Why the need for the bool managementConfigAuthoritative?

JoelSpeed: I think there were several cases we were trying to account for

1. Cluster is a joined cluster, in-cluster config for both, no flags set
2. CMA is running on management, `management-kubeconfig` should be in-cluster, `kubeconfig` should be set to workload cluster
3. CMA is running in workload, `management-kubeconfig` should be set to management cluster, `kubeconfig` should be in-cluster
4. You are running CMA remotely, `kubeconfig` is set to a kubeconfig which then both clients use

I think for number 2 and number 4 to work (this was discussed in CA conversations upstream), we assume that if `kubeconfig != "" && managementKubeconfig == ""` then `kubeconfig` should be used for all clients, the authoritative flag prevents that assumption IIRC?

alexander-demichev: This duplicates autoscaler behaviour 
https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/clusterapi#connecting-cluster-autoscaler-to-cluster-api-management-and-workload-clusters  

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_provider.go#L139

enxebre: The motivation for the bool pattern on the autoscaler was to keep a fallback to previous behaviour by default when --cloud-config is not specified for backward compatibility.
I don't think we need to keep backward compatibility for these flags here since this is a self manage internal component.

Is there a reason we can't just have `--management-cluster-kubeconfig` and `--workload-cluster-kubeconfig`, that seems the most intuitive and should cover all cases?

wdyt?

JoelSpeed: That seems fine to me, will just be one extra flag to know about when developing locally, not a major problem really. This should affect IPI clusters as they will both default to in-cluster config so happy with that if it's acceptable for the HyperShift project

pkg/controller/controller.go
alexander-demichev: it is used for API discovery, following austoscaler pattern here too https://github.com/kubernetes/autoscaler/blob/8dc1afb9a2fd188374620ee0d42358243e087655/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go#L331

main.go
enxebre: @alexander-demichev can we please validate accepted apigroups and fail otherwise?

main.go
```diff
@@ -79,27 +89,30 @@ func main() {
...
-	})
-	if err != nil {
-		klog.Fatal("unable to set up client")
-	}
 	// Prevent the controller from caching node and machine objects.
 	// Stale nodes and machines can cause the approver to not approve certificates
 	// within a timely manner, leading to failed node bootstraps.
-	approverClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
-		Client:      directClient,
+	uncachedManagementClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
```
enxebre, line 95: I'm not sure why we woudn't just use the `mgr.GetAPIReader()` for this clients if we don't want to use the cache.

main.go
```diff
@@ -21,30 +21,35 @@ import (
...
 func main() {
 	var cliConfig string
+	var APIGroup string
+	var managementKubeConfigPath string
+	var workloadKubeConfigPath string
 
 	flagSet := flag.NewFlagSet("cluster-machine-approver", flag.ExitOnError)
 
 	klog.InitFlags(flagSet)
 	flagSet.StringVar(&cliConfig, "config", "", "CLI config")
```
enxebre, line 47: @alexander-demichev can we please validate accepted apigroups and fail otherwise?

main.go
JoelSpeed: IIRC we tried to use the API reader and found there was still an issue with it. I remember being deep into informer territory when debugging this and working out exactly what was happening, will see if I can find my notes.

This isn't the most elegant way to do this, we can actually pass uncached types into the management options, just need to get a PR up to clean that up. Will look at that later.

One advantage of excluding from the cache here is then you can't forget while coding to use the correct client, which I think is advantageous. If we blanket don't cache the object type, this isn't an issue when coding and makes the reconciler logic simpler.

Also, the client constructor in controller runtime can handle this via unstructured objects too so I don't think it's adding more complication on our side

main.go
```diff
@@ -21,30 +21,46 @@ import (
...
+	flagSet.StringVar(&APIGroup, "apigroup", "machine.openshift.io", "API group for machines, defaults to machine.openshift.io")
+	flagSet.StringVar(&managementKubeConfigPath, "management-cluster-kubeconfig", "", "management kubeconfig path,")
+	flagSet.StringVar(&workloadKubeConfigPath, "workload-cluster-kubeconfig", "", "workload kubeconfig path")
+
 	flagSet.Parse(os.Args[1:])
 
+	err := validateAPIGroup(APIGroup)
```
enxebre, line 59: nit: this can be inlined

pkg/controller/controller.go
```diff
@@ -195,7 +199,7 @@ func (m *CertificateApprover) getKubeletCA() *x509.CertPool {
 		Namespace: configNamespace,
 		Name:      kubeletCAConfigMap,
 	}
-	if err := m.Get(context.Background(), key, configMap); err != nil {
+	if err := m.NodeClient.Get(context.Background(), key, configMap); err != nil {
```
enxebre, line 202: this should probably be called workloadClusterClient consistently with the flag and the config, is not only querying nodes.

enxebre:
Please let's make sure to follow up to do some cleanup and review https://github.com/openshift/cluster-machine-approver/pull/114#discussion_r667947787,
https://github.com/openshift/cluster-machine-approver/pull/114#discussion_r668521082 and https://github.com/openshift/cluster-machine-approver/pull/114#discussion_r652834228
/lgtm

main.go
```diff
@@ -38,13 +38,16 @@ import (
...
 	var cliConfig string
+	var APIGroup string
 
 	flagSet := flag.NewFlagSet("cluster-machine-approver", flag.ExitOnError)
 
 	klog.InitFlags(flagSet)
 	flagSet.StringVar(&cliConfig, "config", "", "CLI config")
 	flagSet.Parse(os.Args[1:])
 
+	flagSet.StringVar(&APIGroup, "apigroup", "machine.openshift.io", "API group for machines, defaults to machine.openshift.io")
```
elmiko: i'm curious what you think about making this an environment variable instead of flag. my reason for asking is that we [currently use the `CAPI_GROUP` env variable in the autoscaler](https://github.com/kubernetes/autoscaler/blob/2afb7ab9e1b8e10933e509ed8888a1a0014ff650/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_controller.go#L302) for similar functionality. if we made this an env variable as well, it would be the same mechanism we use for the autoscaler.

not a blocker for me, but i'm curious if we should try to impose some similar behavior here.

elmiko (commented):
this mostly looks good to me, i just had a question about the api group flag and if we shouldn't make it look more like the capi autoscaler implementation.

main.go
```diff
@@ -58,12 +73,20 @@ func main() {
...
+	managementClient, workloadClient, err := createClients(managementConfig, workloadConfig)
+	if err != nil {
+		klog.Fatalf("Can't create clients: %v", err)
+	}
+
 	// Create a new Cmd to provide shared dependencies and start components
 	klog.Info("setting up manager")
-	syncPeriod := 10 * time.Minute
-	mgr, err := manager.New(control.GetConfigOrDie(), manager.Options{
+	mgr, err := manager.New(managementConfig, manager.Options{
```
hanqiuzh, line 88: manager's config here should be workloadConfig? as it will be used for watching csr updates

pkg/machinehandler/machinehandler.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+
+	unstructuredMachineList := &unstructured.UnstructuredList{}
+	unstructuredMachineList.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   m.APIGroup,
+		Kind:    "MachineList",
+		Version: APIVersion,
+	})
+	if err := m.Client.List(m.Ctx, unstructuredMachineList); err != nil {
```
hanqiuzh, line 52: This is a cluster scoped list for machines. @enxebre  do you think if we'll need namespace scoped list?

alexander-demichev: in my case, CAPI and MAPI machines will exist in different namespaces so we want to capture all machines.

JoelSpeed: > CAPI and MAPI machines will exist in different namespaces so we want to capture all machines.

In this scenario, we will still need one for MAPI and one for CAPI right?

I think having the ability to restrict to a particular namespace if required is a reasonable ask, but maybe isn't a blocker for this PR going forward

hanqiuzh: thanks. agree that it isn't a blocker for this PR 👍

main.go
```diff
@@ -76,27 +99,30 @@ func main() {
...
 	// Prevent the controller from caching node and machine objects.
 	// Stale nodes and machines can cause the approver to not approve certificates
 	// within a timely manner, leading to failed node bootstraps.
-	approverClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
-		Client:      directClient,
+	uncachedManagementClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
+		Client:      *managementClient,
+		CacheReader: mgr.GetClient(),
+		UncachedObjects: []client.Object{
+			&corev1.Node{},
```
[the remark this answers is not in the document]
alexander-demichev: you're right, it shouldn’t

main.go
```diff
@@ -76,27 +99,30 @@ func main() {
...
+
+	uncachedWorkloadClient, err := client.NewDelegatingClient(client.NewDelegatingClientInput{
+		Client:      *workloadClient,
 		CacheReader: mgr.GetClient(),
 		UncachedObjects: []client.Object{
-			&machinev1.Machine{},
 			&corev1.Node{},
 		},
+		// CacheUnstructured should be false because we manipulate with unstructured machines
+		CacheUnstructured: false,
```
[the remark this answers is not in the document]
alexander-demichev: right

main.go
JoelSpeed: This comment was resolved without being resolved

JoelSpeed (commented):
/approve

Lets get this to merge, then follow up with the additional feedback

Have had a look at the test failures and they seem to be unrelated. Logs for the CMA indicate it's doing the right thing
/retest-required

PR #116, opened 2021-04-16
Fix registry path in Dockerfile

alexander-demichev:
Fix registry path in Dockerfile, `registry.svc.ci` is deprecated.


elmiko (changes requested):
i think this should update the Dockerfile.rhel as well, it is using the same registry url

Danil-Grigorev:
/approve

Seems passing prow/images

alexander-demichev:
@elmiko Thanks! I added rhel dockerfile

elmiko (approved):
thanks Alex
/lgtm

JoelSpeed:
This will conflict with #104 which is a superset of these changes. It would be preferable to get that PR merged instead

alexander-demichev:
already fixed in other PR

PR #117, opened 2021-04-16
Bug 1940972: Increase cache resync period to 10 min instead of 10 hours

Danil-Grigorev:
Mitigate potential for stale cache and cover edge case for https://bugzilla.redhat.com/show_bug.cgi?id=1940972

This is not a fix for the bug according to discussion on bugzilla, but before the controller-runtime issue is investigated might as well be.

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
I think we should set this to 10 minutes, otherwise LGTM, but yeah, won't be a complete fix for the bug. Do we want to create a separate BZ for this?

I suggest 10 minutes so we don't have to re-establish connections constantly. We _should_ catch every CSR with 15 minutes if we do resync each 10 minute period. If we fail to catch a CSR within its 15 minutes, something wen't very wrong if the resync is at 10 minutes.

Danil-Grigorev:
Changed to 10 minutes @JoelSpeed

PR #118, opened 2021-04-22
Bug 1952282: Ensure serving certs requeued

michaelgugino:
This commit ensures serving certs are requeued in case the node link
controller has not yet updated the machine object.

### DIFF
--- a/pkg/controller/csr_check.go
+++ b/pkg/controller/csr_check.go
@@ -195,7 +195,8 @@ func authorizeCSR(
 	if !ok {
 		klog.Errorf("%v: Serving Cert: No target machine for node %q", req.Name, nodeAsking)
 		//TODO: set annotation/emit event here.
-		return false, nil
+		// Return error so we requeue in case we're racing with node linker.
+		return false, fmt.Errorf("Unable to find machine for node")
 	}
 
 	// SAN checks for both DNS and IPs, e.g.,
--- a/pkg/controller/csr_check_test.go
+++ b/pkg/controller/csr_check_test.go
@@ -644,7 +644,7 @@ func Test_authorizeCSR(t *testing.T) {
 				},
 				csr: goodCSR,
 			},
-			wantErr:   "",
+			wantErr:   "Unable to find machine for node",
 			authorize: false,
 		},
 		{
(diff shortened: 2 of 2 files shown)


lobziik:
/lgtm cancel

Pls, fix unit tests...

pkg/controller/csr_check.go
```diff
@@ -195,7 +195,8 @@ func authorizeCSR(
 	if !ok {
 		klog.Errorf("%v: Serving Cert: No target machine for node %q", req.Name, nodeAsking)
 		//TODO: set annotation/emit event here.
-		return false, nil
+		// Return error so we requeue in case we're racing with node linker.
+		return false, fmt.Errorf("Unable to find machine for node")
```
JoelSpeed, line 199: In a scenario where the Node has come from a Kubelet that isn't managed by Machine API, (think UPI, or BYO bare metal into a vSphere, etc), this will cause the CSR approver to start hot looping trying to approve this CSR which will never be approvable.

I think it is appropriate to requeue yes, but ideally that would be requeue after a few minutes, so we try again within a reasonable time frame, but don't hot loop

JoelSpeed (commented):
As far as I can tell this will put the controller into a hot loop in certain scenarios. I don't think this is an appropriate change in isolation. Can we somehow make this requeue after a few minutes instead?

pkg/controller/csr_check.go
Danil-Grigorev: Should not exponential back-off on requeues manage such scenario?

JoelSpeed: There is exponential backoff yes, but that not only affects this individual resource, but all resources. The default rate limiter used in controller runtime has an overall limit and an individual item limit. So if there were a mixture of Nodes, some backed by Machines, some not, the backoff incurred by the no-Machine Nodes, could have other affects as well.

If we are confident that this won't be an issue, then that's ok, we can proceed as is, but someone should check this

JoelSpeed:
/hold cancel

Have discussed with the team and we think the exponential backoff should be sufficient to keep this safe. We will need to keep an eye on this to check it's not causing issues on mixed platforms and UPI platforms.

elmiko (approved):
thanks Mike!
/lgtm

kborup-redhat:
We are hitting this in 4.12 on Vsphere.

JoelSpeed:
@kborup-redhat Could you please file a new bug with as much relevant information as possible, including a must-gather and we will triage you issue from there

PR #121, opened 2021-05-26
Bug 1956609: Bypass cache when reading Node and Machine objects

JoelSpeed:
We have observed that reading stale data for Machine and Node objects from the cache can cause node bootstrap failures as the approver does not recognise the correct links between CSRs, Machines and Nodes. To prevent reading stale data, only allow the client to return Machine and Node objetcs and lists directly from the API server instead of from the informer cache.


elmiko (approved):
this makes sense to me, i don't suppose there is a meaningful way to test this in unit test? (seems like we'd just be duplicating whatever tests happen in controller-runtime)
/approve

PR #127, opened 2021-07-12
Use controller runtime to initalise client with uncached objects

JoelSpeed:
This should be functionally equivalent looking at the code, this just simplifies the initialisation of the client by letting controller runtime set up the uncached objects rather than having to do this manually ourselves.

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,6 +64,13 @@ func main() {
 	mgr, err := manager.New(control.GetConfigOrDie(), manager.Options{
 		MetricsBindAddress: metricsPort,
 		SyncPeriod:         &syncPeriod,
+		// Prevent the controller from caching node and machine objects.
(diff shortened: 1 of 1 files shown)


Danil-Grigorev:
/lgtm
LGTM!

lobziik:
Do we have test / CI signal for ensure that it works properly?

JoelSpeed:
> Do we have test / CI signal for ensure that it works properly?

This is very hard to test, but in a way yes. If we regress here, then the disruptive tests for etcd quorum restore will start to fail again. The changes that originally motivated this change were affecting that specific test and the test started passing reliably once we fixed the issue with the caching

lobziik:
>If we regress here, then the disruptive tests for etcd quorum restore will start to fail again. 

Sounds ok to me.

/test e2e-aws-disruptive

PR #129, opened 2021-09-13
Bug 2003788: Prevent error loop when a CSR is queued and then approved externally

JoelSpeed:
When adding a CSR to the queue, we filter based on whether the CSR is already approved. We don't however check this again during the reconciliation loop.

In certain scenarios, we can end up in a situation where the CSR is approved externally before we get to it. In this case, the approval call for to the API server fails as we are duplicating the approved condition. When this API call fails, this causes the CSR to requeue and we end up with the CSR in a constant loop of attempting to be approved, but unable for it to actually be approved.

To prevent this, before we reconcile any CSR, we should check that it hasn't already been approved out of band.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -164,6 +164,15 @@ func reconcileLimits(csrName string, machines []machinehandlerpkg.Machine, csrs 
 }
 
 func (m *CertificateApprover) reconcileCSR(csr certificatesv1.CertificateSigningRequest, machines []machinehandlerpkg.Machine) error {
+	// If a CSR is approved after being added to the queue, but before we reconcile it,
+	// it may have already been approved. If it has already been approved, trying to
+	// approve it again will result in an error and cause a loop.
+	// Return early if the CSR has been approved externally.
+	if isApproved(csr) {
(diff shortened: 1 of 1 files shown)


elmiko (commented):
this makes sense to me, and nice description!
/approve

is it worth trying to create an e2e for this situation? (i'm not sure how tough it is to make this happen, but it seems like something we could simulate)

alexander-demichev (approved):
/lgtm

alexander-demichev:
@elmiko We can test it using `envtest` but it might require additional refactoring.

JoelSpeed:
> is it worth trying to create an e2e for this situation? (i'm not sure how tough it is to make this happen, but it seems like something we could simulate)

Very difficult to prove as far as I can tell in an E2E. We need to trigger an event to add the CSR to the queue, and then there is a race, either the test gets there first and approves, in which case we can see the approver hasn't approved, but maybe we can't tell if it's erroring? Or the other way around, in which case we didn't test the mechanism.

I think we may be able to add a unit test but as Alex says, it may require some refactoring.

elmiko:
thanks @alexander-demichev and @joelspeed, i appreciate the extra context. i kinda figured this might be difficult to test.

PR #132, opened 2021-09-30
add commandline flag for status controller

hanqiuzh:
Updates:
- added `--enable-status-controller` to allow users disable/enable status controller
   - default set to true for backward compatibility


hanqiuzh:
/assign @JoelSpeed  @enxebre 

Please review when you have time.  Thanks

PR #133, opened 2021-09-30
add leader election options

hanqiuzh:
Updates:
- added leader election options, same as CCCMO


main.go
```diff
@@ -40,6 +45,18 @@ const (
 	mapiGroup = "machine.openshift.io"
 )
 
+var (
+	// Default leader electrion configuration.
+	leaderElectionConfig = config.LeaderElectionConfiguration{
+		LeaderElect:       true,
+		LeaseDuration:     metav1.Duration{Duration: 137 * time.Second},
```
enxebre, line 52: why did you choose this values?

elmiko:
> what is CCCMO?

https://github.com/openshift/cluster-cloud-controller-manager-operator/ the new operator for managing CCMs on openshift

main.go
```diff
@@ -56,7 +73,11 @@ func main() {
 	flagSet.StringVar(&machineNamespace, "machine-namespace", "", "restrict machine operations to a specific namespace, if not set, all machines will be observed in approval decisions")
 	flagSet.StringVar(&workloadKubeConfigPath, "workload-cluster-kubeconfig", "", "workload kubeconfig path")
 
-	flagSet.Parse(os.Args[1:])
+	// Once all the flags are regitered, switch to pflag
```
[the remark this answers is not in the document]
hanqiuzh: sure [BindLeaderElectionFlags](https://pkg.go.dev/k8s.io/component-base/config/options#BindLeaderElectionFlags) is expecting a pflag as input variable, so I converted the original flags one into pflag type. (same as cccmo)

main.go
hanqiuzh: copied from this commit of cccmo: https://github.com/openshift/cluster-cloud-controller-manager-operator/commit/445bfe3e7d7e78dea80b0dec3f729881880f21db

main.go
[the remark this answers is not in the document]
enxebre: Isn't this exposing more than we need as a flags, e.g ResourceLock?
If I would just expose the ones we need manually, no need for the library.

main.go
```diff
@@ -58,6 +66,12 @@ func main() {
 	flagSet.StringVar(&workloadKubeConfigPath, "workload-cluster-kubeconfig", "", "workload kubeconfig path")
 	flagSet.BoolVar(&disableStatusController, "disable-status-controller", false, "disable status controller that will update the machine-approver clusteroperator status")
 
+	flagSet.BoolVar(&leaderElect, "leader-elect", true, "use leader election when starting the manager.")
```
enxebre, line 69: do we want it true by default?

main.go
```diff
@@ -90,7 +104,14 @@ func main() {
 	// Create a new Cmd to provide shared dependencies and start components
 	klog.Info("setting up manager")
 	mgr, err := manager.New(workloadConfig, manager.Options{
-		MetricsBindAddress: metricsPort,
+		MetricsBindAddress:         metricsPort,
```
JoelSpeed, line 107: Please add `ReleaseOnCancel: true` else this will cause approx 3 minute delays during upgrades

main.go
```diff
@@ -143,7 +164,10 @@ func main() {
 
 	if !disableStatusController {
 		statusController := NewStatusController(mgr.GetConfig())
-		go statusController.Run(1, stop)
+		go func() {
+			<-mgr.Elected()
+			statusController.Run(1, stop)
+		}()
```
JoelSpeed, line 170: Weird that we run this separately, I'll make a note to look into this

JoelSpeed (commented):
One quick change, otherwise LGTM

PR #134, opened 2021-10-01
podsecurity: enforce privileged for openshift-cluster-machine-approver namespace

s-urbaniak:
Starting with OpenShift 4.10 we are introducing PodSecurity admission (https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement).

Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-cluster-machine-approver ` namespace to run privileged pods.

See https://github.com/openshift/enhancements/pull/899 for more details (and excuse the eventual consistency of updates).

/cc @stlaz

### DIFF
--- a/manifests/00-ns.yaml
+++ b/manifests/00-ns.yaml
@@ -12,3 +12,6 @@ metadata:
     openshift.io/run-level: "0"
     # allow openshift-monitoring to look for ServiceMonitor objects in this namespace
     openshift.io/cluster-monitoring: "true" 
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged


enxebre:
> In order not to break control plane workloads this allows workloads in openshift-cluster-machine-approver namespace to run privileged pods.

Why does this need to be privileged? CSRs for master machines during bootstrap are auto approved.

s-urbaniak:
@enxebre your deployment specifies `hostNetwork: true`, see https://github.com/openshift/cluster-machine-approver/blob/e3c5b539effa85db7081f284d909cc85bdec9f8d/manifests/04-deployment.yaml#L27.

The policy violation against baseline states: `host namespaces (hostNetwork=true), hostPort (container \"kube-rbac-proxy\" uses hostPort 9192)`

enxebre:
>The policy violation against baseline states: host namespaces (hostNetwork=true), hostPort (container \"kube-rbac-proxy\" uses hostPort 9192)

Thanks! I question this is actually needed. I can't think of anything that would require the host network for it to work. cc @alexander-demichev @JoelSpeed

deads2k:
> Thanks! I question this is actually needed. I can't think of anything that would require the host network for it to work. cc @alexander-demichev @JoelSpeed

It's an ordering problem.  the machine approver needs to approve a machine so that the kubelet can get a client-cert so that the kubelet can run the CNI plugin daemonset

enxebre:
>It's an ordering problem. the machine approver needs to approve a machine so that the kubelet can get a client-cert so that the kubelet can run the CNI plugin daemonset

mm the control plane csr are automatically approved by a bash script. Then the machine approver runs there and can start interacting with the kas to approve any upcoming CSR for workers, why does it need to be through the host network?

enxebre:
David pointed rebootstrapping as a valid scenario. But I'm still not completely sure, if your control plane certs went expired wouldn’t you need to manually approve the new CSRs anyway, since the kas won’t be able to run the mapprover and so the mapprover won’t be able to operate even with host network in that case?

enxebre:
As a general rule we agree "machine approver needs to move towards being earlier in the chain, not being later.  So host-network is a good idea and static pods are better"

But I question this is actually practical for this component since it’s not responsible for control plane bootstrapping approval and it’d be good to dissect controversial scenarios. That can be done and prioritised separately though.
Defering to cloud team to chime in and merging.
/lgtm

PR #137, opened 2021-10-18
Bug 1860774: Allow fallback to serving cert renewal accounting for egress IPs on SDN

JoelSpeed:
When using the OpenShift Egress IP feature (only available on select platforms, eg vSphere), with SDN, additional IPs are assigned to the network interfaces for the VM. These are then picked up by kubelet and added to the CSR SANs.

In IPI clusters, these IPs are also picked up by Machine API and the IP addresses are listed in the Machine status. During certificate renewals, these IPs are matched and the CSR is approved.

When using IPI clusters, we expect the SANs to match. Because Egress IPs come and go, this isn't reliably the case. To allow this use case, this PR adjusts the CMA so that if the CSRs don't match exactly, because of the IP address list being different, the check will allow the CSR to contain any previously allowed IPs plus any IPs that are listed on the Node's HostSubnet EgressIPs.

This should allow the egress IPs to be moved around by SDN as appropriate and for the CSR certificate renewals to function as expected.

This is not an ideal solution as it doesn't ensure that when Egress IPs are removed, these are then removed from the CSR, but we need to find a solution that will work until CCMs are GA, at which point we believe we can revert this and teach the CCM about egress IP ranges to allow it to exclude the IPs when the CSRs are created

### DIFF
--- a/manifests/01-rbac.yaml
+++ b/manifests/01-rbac.yaml
@@ -134,6 +134,18 @@ rules:
   - machine-approver
   verbs:
   - update
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - networks
+  verbs:
+  - get
+- apiGroups:
+  - network.openshift.io
+  resources:
+  - hostsubnets
+  verbs:
+  - get
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/controller/csr_check.go
+++ b/pkg/controller/csr_check.go
@@ -13,10 +13,13 @@ import (
 	"strings"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
+	networkv1 "github.com/openshift/api/network/v1"
 	machinehandlerpkg "github.com/openshift/cluster-machine-approver/pkg/machinehandler"
 	certificatesv1 "k8s.io/api/certificates/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -34,6 +37,9 @@ const (
 
 	maxMachineClockSkew = 10 * time.Second
 	maxMachineDelta     = 2 * time.Hour
+
+	networkTypeOpenShiftSDN = "OpenShiftSDN"
+	networkClusterName      = "cluster"
 )
 
 var nodeBootstrapperGroups = sets.NewString(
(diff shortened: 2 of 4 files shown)


pkg/controller/csr_check.go
```diff
@@ -159,44 +162,190 @@ func authorizeCSR(
...
-		}
+		approvalErrors = append(approvalErrors, err)
+		klog.Infof("Could not use current serving cert for renewal: %v", err)
+		klog.Infof("Current SAN Values: %v, CSR SAN Values: %v",
+			certSANs(servingCert), csrSANs(csr))
 	}
 
 	// Fall back to the original machine-api based authorization scheme.
 	klog.Infof("Falling back to machine-api authorization for %s", nodeAsking)
+	if err := authorizeServingCertWithMachine(machines, req, nodeAsking, csr); err != nil {
```
lobziik, line 201: maybe wrap it with condition which check machines presence, and do not engage it if no machines there? 
smth like
`if len(machines) > 0 {
... Use Mapi
} else {
... check renewal with egress ips
}`
WDYT?

manifests/01-rbac.yaml
```diff
@@ -121,6 +121,7 @@ rules:
   - config.openshift.io
   resources:
   - clusteroperators
+  - networks
   verbs:
   - get
   - create
```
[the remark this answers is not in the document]
elmiko: +1

pkg/controller/csr_check.go
```diff
@@ -159,44 +162,190 @@ func authorizeCSR(
...
+		return fmt.Errorf("CSR Subject Alternate Name values do not match current certificate")
+	}
+
+	hostSubnet := &networkv1.HostSubnet{}
+	if err := c.Get(context.Background(), client.ObjectKey{Name: nodeName}, hostSubnet); err != nil {
+		return fmt.Errorf("could not fetch hostsubnet: %v", err)
+	}
+
+	allowedIPAddresses := currentCert.IPAddresses
+	for _, ipAddr := range hostSubnet.EgressIPs {
```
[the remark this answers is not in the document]
elmiko: +1

pkg/controller/csr_check.go
```diff
@@ -159,44 +162,202 @@ func authorizeCSR(
...
+	if err := authorizeServingCertWithMachine(machines, req, nodeAsking, csr); err != nil {
+		approvalErrors = append(approvalErrors, err)
+		klog.Infof("Could not use Machine for serving cert authorization: %v", err)
+	} else {
+		// No error means the machine was able to authorize the cert
+		return true, nil
+	}
+
+	network := &configv1.Network{}
+	if err := c.Get(context.Background(), client.ObjectKey{Name: "cluster"}, network); err != nil {
```
enxebre, line 210: can we move "cluster" and "OpenShiftSDN" into constants then move the `get configv1.Network{}` and `network.Status.NetworkType == "OpenShiftSDN"` logic into a unit tested function `isEgress(node)` that it also accounts for https://docs.openshift.com/container-platform/4.6/networking/ovn_kubernetes_network_provider/configuring-egress-ips-ovn.html#nw-egress-ips-node-assignment_configuring-egress-ips-ovn into the discrimitation criteria?

manifests/01-rbac.yaml
```diff
@@ -134,6 +134,23 @@ rules:
   - machine-approver
   verbs:
   - update
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - networks
+  verbs:
+  - get
+  - list
```
enxebre, line 143: why do we need any other than get?

enxebre: shouldn't this be using a regular client rather than a cache for this?

enxebre: https://github.com/openshift/cluster-machine-approver/blob/7b7289a9e6419b9a5ce60cba22b60b9d8ddf3ed5/main.go#L155-L158

JoelSpeed: We could add this as uncached and then avoid the informer, but I don't know if we particularly need to.

In particular, we expect the network object to never change during the lifetime of the cluster and the hostsubnets should be pretty static as well

JoelSpeed:
There is an alternative to this PR, which is to ensure that the nodeip-configuration service runs on all vSphere instances. This may be a better long term solution as it means the CSR only ever contains the one IP picked up by the nodeip-configuration service. We have to assume that the service will pick the correct interface in a scenario where an egress IP is added to the machine even without the API server VIP

pkg/controller/csr_check.go
```diff
@@ -443,6 +543,16 @@ func nodeInternalIP(node *corev1.Node) (string, error) {
 	return "", fmt.Errorf("node %s has no internal addresses", node.Name)
 }
 
+// isEgressEnabled determines whether or not egress IP checks should be enabled.
+func isEgressEnabled(c client.Client) (bool, error) {
+	network := &configv1.Network{}
+	if err := c.Get(context.Background(), client.ObjectKey{Name: networkClusterName}, network); err != nil {
+		return false, fmt.Errorf("could not fetch cluster network: %v", err)
+	}
+
```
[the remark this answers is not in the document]
JoelSpeed: AFAICT this only moves the call, it doesn't reduce the number of calls we are making right? I prefer it as is as there's no need to pass the IPs around

pkg/controller/csr_check_test.go
```diff
@@ -2370,6 +2453,133 @@ func TestAuthorizeServingRenewal(t *testing.T) {
...
+			ca:          []*x509.Certificate{parseCert(t, rootCertGood)},
+			time:        presetTimeCorrect,
+			hostSubnet: &networkv1.HostSubnet{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: testNodeName,
+				},
+			},
+			wantErr: "CSR Subject Alternate Names includes unknown IP addresses",
+		},
+		{
```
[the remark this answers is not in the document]
elmiko: +1

PR #139, opened 2021-11-03
Bug 2019754: Ensure pending CSR count is valid post approval

JoelSpeed:
This is a manual cherry pick of #135

> Currently, we reconcile the limits (which the metrics are sourced from) at the beginning of the reconcile loop.
> 
> We only reconcile when there is a pending CSR.
> 
> So at the beginning of the reconcile loop, there is 1 pending CSR, we set the metric to 1.
> We then approve the CSR, now there are no pending CSRs, so we don't have anything to reconcile.
> With no reconciles happening, we never update the metric value, so it stays at 1.
>
> This PR adjusts so that once we have reconciled, assuming it was successful, we re-reconcile the limits (metrics) to ensure they are still up to date. If we made an approval during the reconcileCSR this should now be accounted for
during the metric collection.
>
> If there are any errors throughout this, we don't need to reconcile the limits as this will be handled when we requeue anyway

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -130,7 +130,16 @@ func (m *CertificateApprover) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	for _, csr := range csrs.Items {
 		if csr.Name == req.Name {
-			return reconcile.Result{}, m.reconcileCSR(csr, machines)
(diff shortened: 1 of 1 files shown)


alexander-demichev (approved):
/lgtm

JoelSpeed:
/approve
On Fedosin's behalf

/label backport-risk-assessed

PR #140, opened 2021-11-16
update to golang 1.17

elmiko:
This change updates the go.mod, makefile, dockerfiles, ci-operator.yaml, and the vendor directory to use golang 1.17 for building. it also adds a minor convenience option to the makefile that allows a user to override the build image.


JoelSpeed:
/lgtm cancel

The unit tests are broken here, this will need further investigation

elmiko:
ack, thanks for the heads up. i'll investigate

elmiko:
seeing a lot of these in the logs
```
Failed to retrieve current serving cert: x509: certificate has expired or is not yet valid: current time 2021-11-30T18:34:23Z is after 2021-11-18T20:12:00Z
```
wonder if we have some bad test data in here

elmiko:
rebased on master with the unit test fix

PR #141, opened 2021-11-17
Bug 2024216: Allow fallback to serving cert renewal accounting for egress IPs on SDN

JoelSpeed:
Manual cherry-pick of #137. 

> When using the OpenShift Egress IP feature (only available on select platforms, eg vSphere), with SDN, additional IPs are assigned to the network interfaces for the VM. These are then picked up by kubelet and added to the CSR SANs.
>
>In IPI clusters, these IPs are also picked up by Machine API and the IP addresses are listed in the Machine status. During certificate renewals, these IPs are matched and the CSR is approved.
>
>When using IPI clusters, we expect the SANs to match. Because Egress IPs come and go, this isn't reliably the case. To allow this use case, this PR adjusts the CMA so that if the CSRs don't match exactly, because of the IP address list being different, the check will allow the CSR to contain any previously allowed IPs plus any IPs that are listed on the Node's HostSubnet EgressIPs.
>
>This should allow the egress IPs to be moved around by SDN as appropriate and for the CSR certificate renewals to function as expected.
>
>This is not an ideal solution as it doesn't ensure that when Egress IPs are removed, these are then removed from the CSR, but we need to find a solution that will work until CCMs are GA, at which point we believe we can revert this and teach the CCM about egress IP ranges to allow it to exclude the IPs when the CSRs are created

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
+	networkv1 "github.com/openshift/api/network/v1"
 	"github.com/openshift/cluster-machine-approver/pkg/controller"
 	"github.com/openshift/cluster-machine-approver/pkg/metrics"
 	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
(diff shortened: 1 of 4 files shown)


elmiko (commented):
thanks Joel, this looks good to me
/approve

given that this was a manual cherry-pick, @lobziik or @enxebre would one of you mind giving a review as well?

miyadav:
pre-merged validation successful more info added to the bug .

/cherry-pick-approved

JoelSpeed:
Adding cherry-pick approved label based on https://github.com/openshift/cluster-machine-approver/pull/141#issuecomment-972832895

PR #143, opened 2021-12-03
Add metric to count old pending CSRs

JoelSpeed:
Currently, we ignore any CSRs that are too old for the machine approver to approve them, based on [this conversation](https://coreos.slack.com/archives/CBZHF4DHC/p1638410837249600) it was asked if we could include a new metric to count the old CSRs as well.

This should mean now that you can determine the three states of CSRs, too new (via a calculation), within the time window, and too old. With the current metrics, there is no way to determine the difference between CSRs that are too new or too old.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -164,10 +164,11 @@ func (m *CertificateApprover) Reconcile(ctx context.Context, req ctrl.Request) (
 func reconcileLimits(csrName string, machines []machinehandlerpkg.Machine, csrs *certificatesv1.CertificateSigningRequestList) bool {
 	maxPending := getMaxPending(machines)
 	atomic.StoreUint32(&MaxPendingCSRs, uint32(maxPending))
-	pending := recentlyPendingCSRs(csrs.Items)
-	atomic.StoreUint32(&PendingCSRs, uint32(pending))
-	if pending > maxPending {
-		klog.Errorf("%v: Pending CSRs: %d; Max pending allowed: %d. Difference between pending CSRs and machines > %v. Ignoring all CSRs as too many recent pending CSRs seen", csrName, pending, maxPending, maxDiffBetweenPendingCSRsAndMachinesCount)
+	recentlyPending, oldPending := pendingCSRsByAge(csrs.Items)
+	atomic.StoreUint32(&PendingCSRs, uint32(recentlyPending))
+	atomic.StoreUint32(&OldPendingCSRs, uint32(oldPending))
+	if recentlyPending > maxPending {
+		klog.Errorf("%v: Pending CSRs: %d; Max pending allowed: %d. Difference between pending CSRs and machines > %v. Ignoring all CSRs as too many recent pending CSRs seen", csrName, recentlyPending, maxPending, maxDiffBetweenPendingCSRsAndMachinesCount)
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


dofinn:
/lgtm

However, would it be possible to add a `requestor` label to this new metric? Then ideally we could RFE this same label into the original `mapi_current_pending_csr`?

I presume OpenShift follows the same metric lifecycling as kubes? https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#metric-lifecycle

elmiko (commented):
i think we should update the metrics doc as well, otherwise this is looking pretty good to me. curious to hear about @damdo 's questions.

JoelSpeed:
> However, would it be possible to add a requestor label to this new metric? Then ideally we could RFE this same label into the original mapi_current_pending_csr?

When you say a label, do you mean a prometheus label, or just some note saying why it where it was requested within the code? Not sure why we would want to expose the requester information within the prometheus for all users to see, have you got a use case I haven't thought of?

> I presume OpenShift follows the same metric lifecycling as kubes? https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#metric-lifecycle

dofinn:
> When you say a label, do you mean a prometheus label, or just some note saying why it where it was requested within the code? Not sure why we would want to expose the requester information within the prometheus for all users to see, have you got a use case I haven't thought of?

@JoelSpeed we have customers that install X that can interact with this, a requestor field enables SRE to see via metric inspect who/what is requesting CSRs

JoelSpeed:
> @JoelSpeed we have customers that install X that can interact with this, a requestor field enables SRE to see via metric inspect who/what is requesting CSRs

JoelSpeed:
> @JoelSpeed we have customers that install X that can interact with this, a requestor field enables SRE to see via metric inspect who/what is requesting CSRs

Ah so you mean, add a label to the metric to show who is requesting the CSRs that haven't been approved? I have concern about that. Typically CSRs are requested by many different entities. Each new label (each new requestor) would create a new time series with Prometheus. Adding a dimension to our metric that is unbounded like this is typically bad form for prometheus as it could cause significant degradation in the performance of Prometheus. For now I'd be keen not to add a new dimension like that

dofinn:
> > @JoelSpeed we have customers that install X that can interact with this, a requestor field enables SRE to see via metric inspect who/what is requesting CSRs
> 
> Ah so you mean, add a label to the metric to show who is requesting the CSRs that haven't been approved? I have concern about that. Typically CSRs are requested by many different entities. Each new label (each new requestor) would create a new time series with Prometheus. Adding a dimension to our metric that is unbounded like this is typically bad form for prometheus as it could cause significant degradation in the performance of Prometheus. For now I'd be keen not to add a new dimension like that

@JoelSpeed this is a fair point. Do we have any data illustrating the current averages of `requestor` ? Ideally as an SRE I want to only care about platform based requests. Would you have any ideas that help SRE address this?

JoelSpeed:
>  Do we have any data illustrating the current averages of requestor

This will be, from the platform at least, around the order of the number of machines. If there are other certificates being created for other purposes, I don't know what these are or the bounds that they may be within. Could end users be using the K8s certificate signer to provide mTLS for pods? In which case it would be much higher than the number of nodes. Unfortunately I don't think this is a very easy number to predict

dofinn:
@JoelSpeed thanks for your thoughts, lets park this and revist if required. Thanks for the great conversation :)

PR #144, opened 2021-12-08
Replace certificates with generated certificactes

JoelSpeed:
This should resolve the expired certificates issues we have been seeing in our tests in a way that we won't see this issue again in the future.

Instead of having generated certs, generate the certs on the fly at the beginning of each test run

### DIFF
--- a/pkg/controller/csr_check_test.go
+++ b/pkg/controller/csr_check_test.go
@@ -1,10 +1,16 @@
 package controller
 
 import (
+	"bytes"
+	"crypto"
+	"crypto/rand"
+	"crypto/rsa"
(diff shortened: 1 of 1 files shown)


damdo:
Good job de-hardcoding the certs, GG.
/lgtm

alexander-demichev (approved):
/approve

JoelSpeed:
/override ci/prow/e2e-upgrade

This change does not affect the payload as the changes are solely within the unit tests. This test failing is a red herring

damdo:
/retitle Bug 2031060: Replace certificates with generated certificates

PR #150, opened 2021-12-17
README.md: Elaborate a bit more on node join

cgwalters:
This came up in the context of https://github.com/redhat-et/microshift

Let's cross-reference to the MCO and specifically the MCS code
which handles the token, since it's important to understand
specifically for OpenShift.


JoelSpeed:
/approve
/skip as this is docs only, we don't need E2Es

elmiko (commented):
my only concern is the deep link into the MCO code, since it might drift over time. but i think this is a solid improvement overall. thanks @lobziik 
/lgtm

lobziik:
/hold cancel
4.11 should be open now

PR #151, opened 2022-01-10
[config] Add disabling node server cert approvals

saifshaikh48:
This PR adds support to disable node server cert auto approvals.

One benefit of this is that the machine approver controller can be stopped from
automatically approving a node's server CSR when it is not the one who approved
the node's client CSR.


saifshaikh48:
@openshift/openshift-team-cloud open for review :rocket:

elmiko:
quick question, is this meant to go into 4.10 or 4.11?

saifshaikh48:
No rush whatsoever. If 4.11 is preferred, that is fine

elmiko:
ok, cool. thanks!

PR #152, opened 2022-01-17
Bug 2028019: Account for large scale simultaneous renewal on UPI clusters

JoelSpeed:
If more than 100 machines need to renew their certificates at once on a UPI cluster today, we exhaust the capacity limit set by the maxPending check. To allow renewal in these cases we need to account for how many Nodes have joined the cluster before we block the approvals

### DIFF
--- a/pkg/controller/csr_check_test.go
+++ b/pkg/controller/csr_check_test.go
@@ -2885,19 +2885,50 @@ func creationTimestamp(delta time.Duration) metav1.Time {
 }
 
 func TestGetMaxPending(t *testing.T) {
-	ml := []machinehandlerpkg.Machine{
+	testCases := []struct {
+		name        string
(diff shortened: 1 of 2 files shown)


pkg/controller/csr_check_test.go
```diff
@@ -2885,19 +2885,50 @@ func creationTimestamp(delta time.Duration) metav1.Time {
...
+			},
+			nodes: []corev1.Node{
+				{},
+				{},
+			},
+			expectedMax: 3 + maxDiffBetweenPendingCSRsAndMachinesCount,
 		},
 		{
-			Status: machinehandlerpkg.MachineStatus{},
+			name: "with more machines than nodes",
```
damdo, line 2908:
```suggestion
			name: "with more nodes than machines",
```

PR #155, opened 2022-02-01
Shrink csr_check_test.go size

lobziik:
This pr contains:
- Moving machines creation to function in Test_authorizeCSR test suite.

 - CRs generation was moved to function, instead of having constants in test file.
For ensure that generated CRs attrs are equal to constant ones this [commit](https://github.com/openshift/cluster-machine-approver/pull/155/commits/eaec2db32f83d6812a2d09cdf5bca98352473b5e) contained test. 
Run results might be checked here: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-machine-approver/155/pull-ci-openshift-cluster-machine-approver-master-unit/1488863004706476032/build-log.txt

### DIFF
--- a/pkg/controller/csr_check_test.go
+++ b/pkg/controller/csr_check_test.go
@@ -64,6 +64,12 @@ SlwIRyqdMvo4Rc6/JUHQ67eXh1Cvy3w6Sg==
 
 var baseTime = time.Date(2020, 11, 19, 0, 0, 0, 0, time.UTC)
 
+// Default CSR generation parameters using across the tests,
+// actual values sets up within the init function
+var defaultOrgs []string
+var defaultIPs []net.IP
+var defaultDNSNames []string
+
 func init() {
 	now = clock.NewFakePassiveClock(baseTime).Now
 	networkv1.AddToScheme(scheme.Scheme)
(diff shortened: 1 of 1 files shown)


pkg/controller/csr_check_test.go
```diff
@@ -422,6 +170,42 @@ func generateCertKeyPair(duration time.Duration, parentCertPEM, parentKeyPEM []b
 	return certOut.Bytes(), keyOut.Bytes(), nil
 }
 
+func createCR(commonName string, organizations []string, ipAddressess []net.IP, dnsNames []string) string {
```
JoelSpeed, line 173: Should this be `createCSR`?

pkg/controller/csr_check_test.go
```diff
@@ -422,6 +170,42 @@ func generateCertKeyPair(duration time.Duration, parentCertPEM, parentKeyPEM []b
...
+	defaultOrganizations := []string{"system:nodes"}
+	defaultIPAdresses := []net.IP{net.ParseIP("127.0.0.1"), net.ParseIP("10.0.0.1")}
+	defaultDnsNames := []string{"node1", "node1.local"}
```
JoelSpeed, line 176: Would it be more obvious what the function is doing if you were required to pass these in rather than defaulting in here? When I see the chain of nils I'm left wondering what should be in there
eg 
```
goodCSR = createCR("system:node:test", defaultOrgs, defaultIPs, deafultDNSNames)
```

JoelSpeed (commented):
This should make things a bit easier to maintain, thanks!

pkg/controller/csr_check_test.go
```diff
@@ -54,283 +59,17 @@ MHcCAQEEIFU6aDkXYwx8YYtXd9AZPIst87daWX49Rzhxsd7D7UwroAoGCCqGSM49
...
 `
 )
 
 var baseTime = time.Date(2020, 11, 19, 0, 0, 0, 0, time.UTC)
 
+// Default CSR generation parameters using across the tests,
+// actual values sets up within the init function
+var defaultOrgs []string
+var defaultIPs []net.IP
+var defaultDNSNames []string
```
JoelSpeed, line 71: Do these need to be globals or can they be defined directly in the init?

damdo:
Aside from Joel's comments, which I saw you addressed, it LGTM.
Good job de-hardcoding these tests and making them even nicer :)
/lgtm

damdo (commented):
Aside from Joel's comments, which I saw you addressed, it LGTM.
Good job de-hardcoding these tests and making them even nicer :)
/lgtm

pkg/controller/csr_check_test.go
lobziik: It can be defined in the init function, i just placed it there for better visibility.

PR #156, opened 2022-02-23
Set API Version based on flag when discovery won't work

JoelSpeed:
In openshift/machine-api-operator#992, we are introducing a new CRD which is in the machine.openshift.io/v1 group. This means that the server preferred version of the API is now `v1` and not `v1beta1. Because Machines don't yet exist in the `v1` group, this means the CMA is currently broken on that PR.

```
E0223 11:12:04.550754       1 controller.go:134] csr-mwwzt: Failed to list machines: no matches for kind "Machine" in version "machine.openshift.io/v1"
```

If we were to merge the PR we would break the CMA across openshift.

To mitigate this, we must allow the version to be set manually and only rely on discovery when the preferred version is not set via a flag.

### DIFF
--- a/main.go
+++ b/main.go
@@ -45,7 +45,8 @@ const (
 
 func main() {
 	var cliConfig string
-	var APIGroup string
+	var apiGroup string
+	var apiVersion string
 	var managementKubeConfigPath string
 	var machineNamespace string
 	var workloadKubeConfigPath string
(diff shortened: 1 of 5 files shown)


alexander-demichev (approved):
/approve

JoelSpeed:
/hold 

@Fedosin is working on an alternative to this

JoelSpeed:
/close

We will use #157 instead

PR #157, opened 2022-02-23
Allow to define several API groups

Fedosin:
Currently approver can work with just one API group, which is defined by --apigroup option. This patch allows to set several API groups, so that the component is able to approve certificates for machines created by various APIs.

For OpenShift it means that we don't have to deploy 2 approvers: one for MAPI and another for CAPI. Instead we can use one instance that manages them all.

Also, using of api-group-version instead of just api-group allows us to define what API version we'd like to use. It's important because In openshift/machine-api-operator#992, we are introducing a new CRD which is in the machine.openshift.io/v1 group. This means that the server preferred version of the API is now v1 and not v1beta1. Because Machines don't yet exist in the v1 group, this means the CMA is currently broken on that PR.

```txt
E0223 11:12:04.550754       1 controller.go:134] csr-mwwzt: Failed to list machines: no matches for kind "Machine" in version "machine.openshift.io/v1"
```

If we were to merge openshift/machine-api-operator#992 we would break the CMA across openshift. To mitigate this, we must allow the version to be set manually as a part of api-group-version, and only rely on discovery when the preferred version is not set there.

Deprecations:
`--apigroup` is deprecated in favor of `--apigroupversion` option

New options:
`--api-group-version` allows to specify both API group and version. This option can be given multiple times.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,17 +17,20 @@ limitations under the License.
 package main
 
 import (
-	"flag"
+	goflag "flag"
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	networkv1 "github.com/openshift/api/network/v1"
 	"github.com/openshift/cluster-machine-approver/pkg/controller"
 	"github.com/openshift/cluster-machine-approver/pkg/metrics"
+	flag "github.com/spf13/pflag"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
@@ -45,7 +48,8 @@ const (
 
 func main() {
 	var cliConfig string
-	var APIGroup string
+	var apiGroupVersions []string
+	var apiGroup string // deprecated
 	var managementKubeConfigPath string
 	var machineNamespace string
 	var workloadKubeConfigPath string
@@ -59,9 +63,11 @@ func main() {
 
 	flagSet := flag.NewFlagSet("cluster-machine-approver", flag.ExitOnError)
 
-	klog.InitFlags(flagSet)
+	klog.InitFlags(nil)
+	flagSet.AddGoFlagSet(goflag.CommandLine)
+
 	flagSet.StringVar(&cliConfig, "config", "", "CLI config")
-	flagSet.StringVar(&APIGroup, "apigroup", "machine.openshift.io", "API group for machines, defaults to machine.openshift.io")
+	flagSet.StringSliceVar(&apiGroupVersions, "api-group-version", nil, "API group and version for machines in format '<group>/<version' or just '<group>'. If version is omitted, it will be set to the latest registered version in the cluster. Defaults to 'machine.openshift.io'. This option can be given multiple times.")
 	flagSet.StringVar(&managementKubeConfigPath, "management-cluster-kubeconfig", "", "management kubeconfig path,")
 	flagSet.StringVar(&machineNamespace, "machine-namespace", "", "restrict machine operations to a specific namespace, if not set, all machines will be observed in approval decisions")
 	flagSet.StringVar(&workloadKubeConfigPath, "workload-cluster-kubeconfig", "", "workload kubeconfig path")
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -59,9 +60,12 @@ func main() {
 
 	flagSet := flag.NewFlagSet("cluster-machine-approver", flag.ExitOnError)
 
-	klog.InitFlags(flagSet)
+	klog.InitFlags(nil)
+	goflag.Parse()
```
JoelSpeed, line 64: Do we need to do this? I think `flag.Parse()` later will handle this

main.go
```diff
@@ -45,7 +46,7 @@ const (
 
 func main() {
 	var cliConfig string
-	var APIGroup string
+	var APIGroups []string
```
JoelSpeed, line 49: Can we rename the variables internally to `apiGroupVersions` to keep this inline with other K8s conventions?

main.go
```diff
@@ -59,9 +60,12 @@ func main() {
...
 	flagSet := flag.NewFlagSet("cluster-machine-approver", flag.ExitOnError)
 
-	klog.InitFlags(flagSet)
+	klog.InitFlags(nil)
+	goflag.Parse()
+	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
+
 	flagSet.StringVar(&cliConfig, "config", "", "CLI config")
-	flagSet.StringVar(&APIGroup, "apigroup", "machine.openshift.io", "API group for machines, defaults to machine.openshift.io")
+	flagSet.StringSliceVar(&APIGroups, "apigroup", []string{"machine.openshift.io"}, "API group for machines, defaults to machine.openshift.io. This option can be given multiple times")
```
JoelSpeed, line 68: How about having a new flag `api-group-version`, then deprecating this old flag? I think it would be better to be more explicit about what we are expecting here

main.go
```diff
@@ -75,8 +79,10 @@ func main() {
 	flagSet.StringVar(&leaderElectResourceNamespace, "leader-elect-resource-namespace", "openshift-cluster-machine-approver", "the namespace in which the leader election resource will be created.")
 	flagSet.Parse(os.Args[1:])
 
-	if err := validateAPIGroup(APIGroup); err != nil {
-		klog.Fatalf(err.Error())
+	for _, apiGroup := range APIGroups {
+		if err := validateAPIGroup(apiGroup); err != nil {
```
JoelSpeed, line 83: This won't work with group version strings unless the function is updated to support group versions

pkg/controller/controller.go
```diff
@@ -40,8 +40,8 @@ type CertificateApprover struct {
 	MachineRestCfg   *rest.Config
 	MachineNamespace string
 
-	Config   ClusterMachineApproverConfig
-	APIGroup string
+	Config    ClusterMachineApproverConfig
+	APIGroups []string
```
JoelSpeed, line 44: This should be group versions, what about using a [schema.GroupVersion](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#GroupVersion) as well, this would make it easier to grab the group versions out, and, we can use the parse function in the main to validate the flag options
```suggestion
	APIGroupVersions []schem.GroupVersion
```

pkg/controller/controller.go
```diff
@@ -125,14 +125,18 @@ func (m *CertificateApprover) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	machines, err := machineHandler.ListMachines()
-	if err != nil {
-		klog.Errorf("%v: Failed to list machines: %v", req.Name, err)
-		return reconcile.Result{}, fmt.Errorf("Failed to list machines: %w", err)
+	var machines []machinehandlerpkg.Machine
+
+	for _, apiGroup := range m.APIGroups {
+		newMachines, err := machineHandler.ListMachines(apiGroup)
+		if err != nil {
+			klog.Errorf("%v: Failed to list machines: %v", req.Name, err)
```
JoelSpeed, line 136: Would be good to add the group version into this log message

JoelSpeed (commented):
I understood from slack that we would be adding versioning to this, but it doesn't seem to be handled yet? What's the plan here?

main.go
```diff
@@ -75,8 +80,42 @@ func main() {
...
+	if apiGroup != "" && len(apiGroupVersions) > 0 {
+		klog.Fatal("Cannot set both --apigroup and --apigroupversion options together.")
+	}
+
+	var parsedAPIGroupVersions []schema.GroupVersion
+
+	if len(apiGroupVersions) > 0 {
+		// Parsing API Group Versions
+		for _, apiGroupVersion := range apiGroupVersions {
+			parsedAPIGroupVersion, err := schema.ParseGroupVersion(apiGroupVersion)
```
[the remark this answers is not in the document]
JoelSpeed: If that is our intended behaviour, please document it within the flag

enxebre:
Can you please clarify on the desc what's the use case that motivates this change?

JoelSpeed:
> Can you please clarify on the desc what's the use case that motivates this change?

The versioning motivation is because of introducing new v1 resources within machine.openshift.io, I had originally raised #156 

The multiple groups is because, at the moment, having a CAPI and MAPI CSR approver causes lots of errors in the logs. Half the CSRs cannot be approved by one or the other, so if we want to reduce the errors, we need to just have a single CMA within the cluster that can handle both types

main.go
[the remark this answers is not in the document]
Fedosin: Yep, I updated the patch - now a string without / means the group, not the version. It is described in the description

JoelSpeed:
I'm happy with the code changes, but would like to check with @enxebre about his concerns before we merge

enxebre:
ok thanks I see, why did we put this APIs in v1 https://github.com/openshift/api/tree/master/machine/v1 rather than being with the core types in v1beta1?

JoelSpeed:
@Fedosin can you update the PR so that the CMA deployed on HA clusters has the correct flag? That file is in the same folder

elmiko:
> ok thanks I see, why did we put this APIs in v1 https://github.com/openshift/api/tree/master/machine/v1 rather than being with the core types in v1beta1?

we've been asked to not bring new API objects into the v1beta1 version. that's why we created the alibaba providerspec in the v1, and also for the placement groups as well. as i understand it, the api team does not want to introduce new resources in the v1beta1 version, and i gather this has to do with our support life cycle around APIs that are not v1.

you can see David's explanation (which is much better than mine) here: https://github.com/openshift/api/pull/1045#issuecomment-986891061

main.go
```diff
@@ -75,8 +81,42 @@ func main() {
 	flagSet.StringVar(&leaderElectResourceNamespace, "leader-elect-resource-namespace", "openshift-cluster-machine-approver", "the namespace in which the leader election resource will be created.")
 	flagSet.Parse(os.Args[1:])
 
-	if err := validateAPIGroup(APIGroup); err != nil {
-		klog.Fatalf(err.Error())
+	// Deprecated options
+	flagSet.StringVar(&apiGroup, "apigroup", "machine.openshift.io", "API group for machines")
```
JoelSpeed, line 85: This should now have an empty value else you can't set the new flag
```suggestion
	flagSet.StringVar(&apiGroup, "apigroup", "", "API group for machines")
```

JoelSpeed (commented):
/lgtm cancel

This is broken right now as both flags are set (the old flag needs its default removed)

PR #158, opened 2022-03-07
Bug 1978303: update approve condition logic

elmiko:
This change updates the approval condition logic to detect when the
condition already exists in the CSR record. If the condition already
exists, and has a different state than is desired, the condition will be
updated. If the condition does not exist, it will be created and added
to the CSR record.

When the condition does not need to be updated, the approve logic will
now skip sending an update to the API server.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -270,21 +270,49 @@ func (m *CertificateApprover) getKubeletCA() *x509.CertPool {
 }
 
 func approve(rest *rest.Config, csr *certificatesv1.CertificateSigningRequest) error {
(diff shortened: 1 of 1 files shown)


elmiko:
@JoelSpeed ptal, the function in the mao is slightly different (more generic), so i've modified it for this application. i did look at library-go to see about possibly creating a common function for this, but i think we will need to do some planning if we want to realize that as we need to analyze the various usages and build a function that can work in multiple places and for multiple condition types.

JoelSpeed (commented):
/approve

damdo (commented):
/lgtm

PR #160, opened 2022-03-15
Bug 2047702: Reconcile CSRs approved by other controllers

RadekManak:
Fixes mapi_current_pending_csr metric being stuck at 1
after CSR is rejected by CMA and subsequently approved by another
controller

Not sure if using condition message in logic is bad, but storing approving controller in there avoids running redundant reconciles.

Why the bug occurs:
There are two types of CSRs, client CSR and server CSR. CMA handles both for initial approval. But only the server certificate during renewal. What happens is we get client CSR to reconcile, recognize that it is not for us to approve and leave it pending. Increasing mapi_current_pending_csr metric.
After that kube-controller-manager approves the CSR but we ignore the certificate because it is already approved leaving the metric unchanged.

How this PR fixes the issue:
We run reconcile on CSRs if they have been approved by another controller in the last 30 seconds. This allows us to reconcile metrics on all changes in the cluster making the metric accurate.

When reconciling approved CSRs we gather information (CSRs, machines, nodes) required for updating metrics, update the metrics, log that CSR is approved and exit.

How tested:
1. Create an invalid server certificate so that it won't get automatically approved. 
https://gist.github.com/RadekManak/e9a005a41617e7f2cf06ff4bec824773
`oc apply -f csr.yaml`
2. mapi_current_pending_csr is set to 1 because CSR did not get approved
3. Manually approve using `oc adm certificate approve csr-test`
4. Observe that mapi_current_pending_csr is now correctly reporting 0 on this version.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=2047702

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	configNamespace    = "openshift-config-managed"
-	kubeletCAConfigMap = "csr-controller-ca"
+	configNamespace            = "openshift-config-managed"
+	kubeletCAConfigMap         = "csr-controller-ca"
+	csrConditionApproveMessage = "This CSR was approved by the Node CSR Approver (cluster-machine-approver)"
 )
 
 // MachineApproverReconciler reconciles a machine-approver  object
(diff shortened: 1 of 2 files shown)


JoelSpeed:
Can you please update the description to explain exactly why this fixes the bug. If I understand correctly, the idea is that we will reconcile any CSR that is either unapproved, or approved by a different controller. In the case it is approved by another controller, what happens when the reconcile starts? We update the metrics and?

Can you also please add some notes about how you've tested this

alexander-demichev (approved):
/lgtm

PR #166, opened 2022-06-02
Support Azure ephemeral OS Disk machine redeploys

darthhexx:
Creating an Azure ephemeral OS disk node works as expected in OCP 4.10, however when a machine is redeployed it will never rejoin the cluster due to the node already existing, as well as the nodeRef and potentially failing the creation time test, depending on how soon the node is redeployed after creation.

This is due to the fact that, when an ephemeral OS disk node is redeployed, the OS is re-provisioned during the process. This limitation is listed in the [Azure ephemeral OS disk docs](https://docs.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks).

Due to the fact that we use redeploy actions to support clusters in ARO extensively, we are currently blocked from offering this feature on the platform until this issue is mitigated.

The approach taken in this PR does result in a circumvention of some validations, but keen to iterate to get the concept merged.


darthhexx:
This is not a direction that is going to be explored and we'll instead move to a machine delete approach going forward.

PR #170, opened 2022-08-25
ignore case when checking csr hostnames

AlexVulaj:
Related to OSD-12989.  The machine approver doesn't recognize hostnames that use capital letters as valid even though DNS is case-insensitive.  This change will allow the machine approver to properly recognize those hostnames.

### DIFF
--- a/pkg/controller/csr_check.go
+++ b/pkg/controller/csr_check.go
@@ -373,9 +373,9 @@ func authorizeServingCertWithMachine(machines []machinehandlerpkg.Machine, req *
 		var attemptedAddresses []string
 		var foundSan bool
 		for _, addr := range targetMachine.Status.Addresses {
-			switch corev1.NodeAddressType(addr.Type) {
+			switch addr.Type {
 			case corev1.NodeInternalDNS, corev1.NodeExternalDNS, corev1.NodeHostName:
-				if san == addr.Address {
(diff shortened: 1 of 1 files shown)


pkg/controller/csr_check.go
```diff
@@ -373,9 +373,9 @@ func authorizeServingCertWithMachine(machines []machinehandlerpkg.Machine, req *
 		var attemptedAddresses []string
 		var foundSan bool
 		for _, addr := range targetMachine.Status.Addresses {
-			switch corev1.NodeAddressType(addr.Type) {
+			switch addr.Type {
```
AlexVulaj, line 376: This was an unrelated clean up I found while stepping through the code - unnecessary conversion that can be undone.  Let me know if there's any reason we should leave this as it was.

JoelSpeed:
This seems reasonable
/approve

damdo (commented):
Thanks for submitting this change @AlexVulaj
/lgtm

damdo (commented):
/hold
for QE review

damdo:
Thanks @AlexVulaj for submitting this change
/lgtm

damdo:
Thanks for reviewing @huali9

/unhold

RadekManak:
/retitle OCPBUGS-676: ignore case when checking csr hostnames

PR #173, opened 2022-10-26
OCPBUGS-2621: Update capi related CMA deployment ports

lobziik:
In case of single node installations two CMA deployments tries to use same set of ports for metrics.
Set another ports for CAPI related CMA deployment.

### DIFF
--- a/manifests/04-deployment-capi.yaml
+++ b/manifests/04-deployment-capi.yaml
@@ -29,8 +29,8 @@ spec:
       serviceAccountName: machine-approver-sa
       containers:
       - args:
-        - --secure-listen-address=0.0.0.0:9192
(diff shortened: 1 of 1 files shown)


lobziik:
https://github.com/openshift/enhancements/pull/1272 - related pr to the port registry document.

manifests/04-deployment-capi.yaml
```diff
@@ -82,7 +82,7 @@ spec:
         - name: RELEASE_VERSION
           value: "0.0.1-snapshot"
         - name: METRICS_PORT
```
JoelSpeed, line 84: Is this observed by the CMA, don't need to set a flag or anything?

PR #175, opened 2022-12-14
OCPBUGS-3762: update code for isNodeClientCert

Fedosin:
Originally isNodeClientCert was copied from [k8s repo](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/certificates/approver/sarapprove.go), and hasn't been modified after that. Unfortunately, we see a lot of failures in this function, so it's better to sync the code with k8s once again.

```
csr_check.go:157] csr-5698l: CSR does not appear to be client csr
```

### DIFF
--- a/pkg/controller/node_client.go
+++ b/pkg/controller/node_client.go
@@ -2,51 +2,116 @@ package controller
 
 import (
 	"crypto/x509"
+	"fmt"
 	"reflect"
 	"strings"
 
-	certificatesv1 "k8s.io/api/certificates/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	capi "k8s.io/api/certificates/v1"
+	"k8s.io/klog/v2"
 )
 
 // copied from k8s.io/kubernetes/pkg/controller/certificates/approver/sarapprove.go
 
-func hasExactUsages(csr *certificatesv1.CertificateSigningRequest, usages []certificatesv1.KeyUsage) bool {
-	if len(usages) != len(csr.Spec.Usages) {
+type KeyUsage string
+
+const (
+	UsageSigning           KeyUsage = "signing"
+	UsageDigitalSignature  KeyUsage = "digital signature"
+	UsageContentCommitment KeyUsage = "content commitment"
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -5,74 +5,73 @@ go 1.18
...
 	github.com/openshift/library-go v0.0.0-20220920133651-093893cf326b
-	github.com/prometheus/client_golang v1.12.2
+	github.com/prometheus/client_golang v1.14.0
 	github.com/spf13/pflag v1.0.5
 	k8s.io/api v0.25.1
 	k8s.io/apimachinery v0.25.1
 	k8s.io/client-go v0.25.1
 	k8s.io/klog/v2 v2.80.1
-	k8s.io/utils v0.0.0-20220823124924-e9cbc92d1a73
+	k8s.io/kubernetes v1.26.0
```
JoelSpeed, line 18: Do we really want to import Kubernetes? It's a massive dependency with a long list of other dependencies which might make our own dependency management harder.

What's the diff between the util itself and the code we have? Should we just copy/paste it over as we did before?

JoelSpeed:
Please make notes either here or in the bug about why this is the appropriate fix for the bug.
Glancing at the bug now, there's nothing that explains what the actual issue faced by the CI runs is, nor why this might fix it, please write up your logic/discoveries that lead you to this fix

go.mod
Fedosin: ack, it makes sense. I'll update the code.

pkg/controller/node_client.go
```diff
@@ -2,51 +2,116 @@ package controller
...
+	kubeletClientRequiredUsagesNoRSA = sets.NewString(
+		string(UsageDigitalSignature),
+		string(UsageClientAuth),
+	)
+	kubeletClientRequiredUsages = sets.NewString(
+		string(UsageDigitalSignature),
+		string(UsageKeyEncipherment),
+		string(UsageClientAuth),
+	)
+)
```
JoelSpeed, line 65: How many of these are actually needed? Can we trim this to just the ones we actually need?

PR #183, opened 2023-03-30
Patch CSR name onto Machine annotations

dulek:
Sometimes it may happen that it takes Machine controller some time to reconcile the Machine and update object with current state while cloud provider already did so. In such situations automatic approval of the CSR may fail because Machine object Status is not up to date.

In order to force reconciliation of the Machine object this commit implements adding an annotation with the CSR name to the Machine object when CSR verification fails.

### DIFF
--- a/manifests/01-rbac.yaml
+++ b/manifests/01-rbac.yaml
@@ -135,6 +135,7 @@ rules:
   - get
   - list
   - watch
+  - patch
 - apiGroups:
   - ""
   resources:
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -156,6 +156,13 @@ func (m *CertificateApprover) Reconcile(ctx context.Context, req ctrl.Request) (
 	for _, csr := range csrs.Items {
 		if csr.Name == req.Name {
 			if err := m.reconcileCSR(csr, machines); err != nil {
+				// Annotate the machine with its CSR in order to force reconciliation so that we'll make sure it's up to
+				// date for the next try to approve it.
+				err = m.patchCSRNameToMachine(csr, machines, machineHandler)
+				if err != nil {
+					klog.Warningf("couldn't patch machine with CSR %s: %s, it will not be bumped, ignoring", csr.Name, err)
+					// Ignoring error on purpose, if we couldn't patch, then whatever.
+				}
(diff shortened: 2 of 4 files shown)


dulek:
/retest

Failures seem unrelated.

mdbooth:
> One concern I have is that this is creating an annotation based API, any reason not to create a condition on the Machine's condition list? Operators working on conditions should behave well and only operate on their own condition/not interfere with others, especially easily if we can use SSA for this

This fell through the cracks and I'm just seeing it. @JoelSpeed would this condition be added by, and managed by the CSR approver? Is that an accepted pattern?

dulek:
> > This fell through the cracks and I'm just seeing it. @JoelSpeed would this condition be added by, and managed by the CSR approver? Is that an accepted pattern?
> 
> The conditions list is considered a map in Kube with the condition type as the key. Different controllers may add different conditions to an object so long as they do not clash on the type field. This is quite a common pattern, though it assumes everyone is a good actor. You can use server side apply for this to avoid having to worry about conflicting with others

Yes, I was postponing working on this one as bug has a simple workaround, but we got to get back to it eventually.

@JoelSpeed: Can you point me to an example on how to use SSA in golang client? I had hard time trying to figure it out myself.

JoelSpeed:
I don't have an example to hand, but the gist of it is:
* Extract the status (since that's what we care about) using [https://pkg.go.dev/k8s.io/client-go@v0.27.3/applyconfigurations/certificates/v1#ExtractCertificateSigningRequestStatus](https://pkg.go.dev/k8s.io/client-go@v0.27.3/applyconfigurations/certificates/v1#ExtractCertificateSigningRequestStatus)
* Update the object that comes back form this (not a CSR, but an apply configuration version of it)
* Use the CSR client to [apply this object via ApplyStatus](https://github.com/kubernetes/client-go/blob/v0.27.3/kubernetes/typed/certificates/v1/certificatesigningrequest.go#L43). The field manager should be unique to the CSR approver.

To be honest though, for this case, there aren't many operators acting on CSRs. I would probably just do it the old fashioned way, fetch all of them, make sure you only modify your one, update it

mdbooth:
@dulek Sounds like we're not going to do this, but do the condition thing instead?

PR #184, opened 2023-03-31
OCPBUGS-11225: approver: fix ECDSA approvals in 1.27

rphillips:
Relax the validation usages needed for ECDSA CSRs.

ref: https://github.com/kubernetes/kubernetes/pull/111660

### DIFF
--- a/pkg/controller/csr_check.go
+++ b/pkg/controller/csr_check.go
@@ -77,25 +77,31 @@ func validateCSRContents(req *certificatesv1.CertificateSigningRequest, csr *x50
 		return "", fmt.Errorf("%q not in %q and %q", groupSet, "system:authenticated", nodeGroup)
 	}
 
+	validationUsageSetLegacy := []string{
+		string(certificatesv1.UsageDigitalSignature),
+		string(certificatesv1.UsageKeyEncipherment),
+		string(certificatesv1.UsageServerAuth),
+	}
+	validationUsageSet := []string{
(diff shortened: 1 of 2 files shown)


soltysh (approved):
/lgtm

pkg/controller/csr_check.go
```diff
@@ -77,25 +77,38 @@ func validateCSRContents(req *certificatesv1.CertificateSigningRequest, csr *x50
 		return "", fmt.Errorf("%q not in %q and %q", groupSet, "system:authenticated", nodeGroup)
 	}
 
+	var validationUsageSet []string
+
+	switch csr.PublicKeyAlgorithm {
+	// ECDSA keys only require UsageDigitalSignature and UsageServerAuth
+	case x509.ECDSA:
```
JoelSpeed, line 84: In the renewal controller upstream they are allowing both usage sets for ECDSA, which I think is for upgrades, we will need to work out a way to refactor this to allow both as this will break 4.13 compat as it is right now

JoelSpeed:
Matches upstream logic
/lgtm

pkg/controller/csr_check.go
rphillips: Fixed

JoelSpeed:
/override ci/prow/e2e-aws-capi-techpreview 

Confident this isn't going to be the source of issues for the techpreview jobs if it is working for the main set of jobs, so going to continue this

PR #194, opened 2023-04-26
Go 1.20 bump with fixed unit tests

RadekManak:
Go 1.20 adds additional wrapping to the unknown authority error. This fails our test.

Fixes the test and bumps go.mod go version, because the test only works for go 1.20 and above now.


RadekManak:
/retitle OCPBUGS-10171:  Go 1.20 bump with fixed unit tests
supersedes #193

PR #197, opened 2023-08-07
Ignore if machines are not found

qJkee:


pkg/machinehandler/machinehandler.go
```diff
@@ -66,7 +67,9 @@ func (m *MachineHandler) ListMachines(apiGroupVersion schema.GroupVersion) ([]Ma
 		listOpts = append(listOpts, client.InNamespace(m.Namespace))
 	}
 	if err := m.Client.List(m.Ctx, unstructuredMachineList, listOpts...); err != nil {
-		return nil, err
+		if !k8serrors.IsNotFound(err) {
```
JoelSpeed, line 70: And if it is a not found? Should we log and return early?

pkg/machinehandler/machinehandler.go
```diff
@@ -66,6 +67,12 @@ func (m *MachineHandler) ListMachines(apiGroupVersion schema.GroupVersion) ([]Ma
 		listOpts = append(listOpts, client.InNamespace(m.Namespace))
 	}
 	if err := m.Client.List(m.Ctx, unstructuredMachineList, listOpts...); err != nil {
+		// Sometimes when MachineAPI disabled k8s returns StatusReasonUnknown for the List
+		// request, that's why we check this reason as well and return nil because there
+		// are no machines present
+		if k8serrors.IsNotFound(err) || k8serrors.ReasonForError(err) == metav1.StatusReasonUnknown {
```
JoelSpeed, line 73: This is too broad, reason unknown could happen for too many reasons for this to be safe IMO.

Please add a discovery client check on startup that determines if the Machine API is installed or not, and then use that to determine if the Machine components should be used.

I'm not happy with the way this is just relying on errors.

I left a comment on the previous PR with a link to an example of using the discovery client, so you should have some prior art to base on

JoelSpeed:
This PR needs to be attached to the relevant bug and have an appropriate PR description added please

PR #207, opened 2023-11-07
[OCPBUGS-21594] Add a metric of pending CSRs that will be handle by machine approver.

bergmannf:
Instead of just storing all pending CSRs, add a new metric that only holds the number of pending CSRs that are node CSRs.

I didn't want to overwrite the old metric, as that would disconnect the metric name from what's actually stored in the metric.


bergmannf:
Seems to be replaced by #208

PR #208, opened 2023-11-07
OCPBUGS-21594: Filter non node CSRs in metrics

RadekManak:
CSRs not created by node bootstrapper are correctly filtered out in the reconciler, but this filtering was not performed when reporting metrics. 

This PR adds the necessary filtering and clarifies what the metrics report.


bergmannf:
Ah thanks for opening this one - I'm closing #207 as this seems to perform a similar fix, but actually replaces the metrics (I wasn't sure that's desired).

RadekManak:
Updated to also count server CSRs as we also reconcile them.

PR #231, opened 2024-05-17
OCPBUGS-33644: check for machine crd before listing machines

qJkee:
Avoid fetching machines from api when machines api is not present in the cluster

### DIFF
--- a/pkg/machinehandler/machinehandler.go
+++ b/pkg/machinehandler/machinehandler.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	corev1 "k8s.io/api/core/v1"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


pkg/machinehandler/machinehandler.go
```diff
@@ -126,6 +140,20 @@ func (m *MachineHandler) getAPIGroupPreferredVersion(apiGroup string) (string, e
 	return "", ErrApiGroupNotFound
 }
 
+func isMachineCRDPresent(ctx context.Context, cl client.Client) (bool, error) {
+	unstructuredCRD := &unstructured.Unstructured{}
+	unstructuredCRD.SetGroupVersionKind(v1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
+
+	err := cl.Get(ctx, client.ObjectKey{Name: "machines.machine.openshift.io"}, unstructuredCRD)
```
JoelSpeed, line 147: I think this needs to come after the API group preferred version check. Also, the rest of the checks are not specifically referring to the Machine API group, since the approver can work with different API groups

JoelSpeed, line 147: You could also use the `ServerResourcesForGroup` function on the dynamic client to work this out. It looks for the API Group and works out what resources are installed, might be neater than checking for the CRD.

It's a very easy extension to what we already have

manifests/01-rbac.yaml
```diff
@@ -198,6 +198,14 @@ rules:
...
   verbs:
   - get
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  resourceNames:
+  - machines.machine.openshift.io
+  verbs:
+  - get
```
JoelSpeed, line 208: We don't need this anymore right?

pkg/machinehandler/machinehandler_test.go
```diff
@@ -177,5 +190,14 @@ func Test_authorizeCSR(t *testing.T) {
...
+}
 
+var machineCRD = &unstructured.Unstructured{
+	Object: map[string]interface{}{
+		"apiVersion": v1.SchemeGroupVersion.String(),
+		"kind":       "CustomResourceDefinition",
+		"metadata": map[string]interface{}{
+			"name": "machines.machine.openshift.io",
+		},
+	},
```
JoelSpeed, line 202: Do we need this still?

pkg/machinehandler/machinehandler_test.go
```diff
@@ -67,10 +68,22 @@ func (f fakeMachineRoundTripper) RoundTrip(req *http.Request) (*http.Response, e
 				  }
 			]
 		}`
+	} else {
```
JoelSpeed, line 71: You should use the path here, not just assume all remaining requests hit this point, the path should be `/apis/<group>/<version>` so you'll need to account for both versions tested in this test suite

PR #237, opened 2024-08-12
OCPBUGS-36871: Client internal DNS checks should be case insensitive

JoelSpeed:
DNS names are case insensitive. Kubelet automatically lower cases names, but, our actuators may not (especially on AWS). In this case, the client certificates cannot be approved, because we currently look for an exact match between the node name and the DNS name.


elmiko (commented):
/approve

PR #248, opened 2024-12-05
OCPBUGS-45306: Client internal DNS checks should ignore trailing dot

huali9:
Users may desire to configure their AWS DHCP Option Set with a custom domain name, that contains a trailing dot, but a trailing dot is not permitted in a Kubernetes object name per the DNS 1123 subdomain name standard, so met the bug [OCPBUGS-45306](https://issues.redhat.com/browse/OCPBUGS-45306). We fixed that in the pr https://github.com/openshift/machine-config-operator/pull/4729 to ensure that we trim any trailing dot prior to passing the hostname over to kubelet. 
But a new issue is the client certificates cannot be approved, because we currently look for an exact match (only ignore case) between the node name and the DNS name, and in this case the node name without the trailing dot, but the DNS name with the trailing dot. So update it to ignore the trailing dot when matching.

### DIFF
--- a/pkg/controller/csr_check_test.go
+++ b/pkg/controller/csr_check_test.go
@@ -859,6 +859,33 @@ func Test_authorizeCSR(t *testing.T) {
 			authorize: true,
 		},
 		{
+			name: "client good with trailing dot in DNS",
+			args: args{
+				machines: []machinehandlerpkg.Machine{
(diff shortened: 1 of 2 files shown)


shellyyang1989:
cc @JoelSpeed for review. Thank you.

JoelSpeed:
/cherry-pick release-4.18 release-4.17 release-4.16 release-4.15 release-4.14

PR #253, opened 2024-12-11
OCPBUGS-45306: Ensure trailing dots on DNS names do not block serving cert auth

JoelSpeed:
This is a follow up to https://github.com/openshift/cluster-machine-approver/pull/248, which fixed the issue for client certificates. However, we also need to fix this for serving certificates so that we can access things like logs or oc debug on the node.


JoelSpeed:
/cherry-pick release-4.18 release-4.17 release-4.16 release-4.15 release-4.14

PR #269, opened 2025-05-01
OCPBUGS-19856: Improve log message when searching for node by InternalDNS

mdbooth:
- [x] Merge https://github.com/openshift/cluster-machine-approver/pull/270 to suppress Snyk false positives

PR #270, opened 2025-05-05
No-Jira: ignore test file csr_check_test.go

miyadav:
Checking if ignoring the test file , will remove snyk failure .

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -4,3 +4,4 @@
 exclude:
   global:
     - "vendor/**"
+    - "pkg/controller/csr_check_test.go"


miyadav:
@JoelSpeed @damdo PTAL when time permits.

mdbooth:
I've tried adding a test commit to https://github.com/openshift/cluster-machine-approver/pull/269 which ignores the specific false positives rather than generally excluding the file. The `.snyk` file is quite poorly documented, though, so not sure it will work.

miyadav:
> I've tried adding a test commit to #269 which ignores the specific false positives rather than generally excluding the file. The `.snyk` file is quite poorly documented, though, so not sure it will work.

acked will keep eye on that as well, thanks.

mdbooth:
> > I've tried adding a test commit to #269 which ignores the specific false positives rather than generally excluding the file. The `.snyk` file is quite poorly documented, though, so not sure it will work.
> 
> acked will keep eye on that as well, thanks.

Didn't work, and can't be bothered to reverse engineer the file format. Lets go with this. I don't see these reports as very high value, anyway.

JoelSpeed:
You can also ignore things through the Snyk dashboard, though I've found logging into it can be very hit and miss

mdbooth:
Can I approve this, I wonder?

/approve
/lgtm

mdbooth:
Nope. @JoelSpeed could you please bless this appropriately?

miyadav:
@JoelSpeed , the problem is it is showing like 40+ such [vulnerabilities](https://app.snyk.io/org/openshift-ci-internal/project/deb648b2-34ad-45f3-b9cd-a7607016b75c/history/8f996f48-e25d-4fd7-ba9f-4876cde3a1a4?fromGitHubAuth=true)(low) for this file , we have to ignore them one by one. .

mdbooth:
This needs https://github.com/openshift/cluster-capi-operator/pull/289 to merge before e2e-aws-capi-techpreview will pass.

PR #309, opened 2026-07-03
OCPBUGS-57437: Always validate EgressIPs

mdbooth:
EgressIPs manifest as additional IP addresses on a Node. They are dynamic, so can be added and removed. Even though they cannot be used to serve incoming traffic, CCMs cannot currently distinguish EgressIPs and they will typically be added to a Node's Addresses. Consequently kubelet will typically request a CSR containing EgressIPs when they are present on a Node.

We had some existing code to make the approver aware of OpenShiftSDN egress IPs, but no equivalent for OVN Kubernetes.

This change fixes 2 aspects of validating a Node with EgressIPs:
* Extend support for EgressIP matching to OVN Kubernetes
* Ensure that a Node cannot continue to request an EgressIP which it no longer has

For the first, we also fetch OVNKubernetes EgressIP based on the infrastructure's networkType.

For the second, we can only do this for IPI clusters as we don't know the complete set of permitted address in a UPI cluster. To enable this for IPI clusters, we deterministically split the authorisation path chosen based on the presence of a Machine object for a Node. We no longer permit fall-through authentication. A side benefit of this is that the error messages reported on authentication failures become shorter. However, note from the test changes that this doesn't result in any additional failure modes.

## Summary by CodeRabbit

* **New Features**
  * CSR approval now supports OVN-Kubernetes EgressIP resources when validating serving certificates.
  * Serving-certificate SAN validation accepts assigned node egress IPs alongside existing certificate addresses.
  * Certificate requests can be authorized using machine identity or validated kubelet certificate renewal.

* **Bug Fixes**
  * Improved CSR authorization error handling by distinguishing validation failures from retryable issues.
  * Unauthorized certificate requests are now ignored without unnecessary retries.

* **Chores**
  * Updated supporting dependencies and access permissions for EgressIP resources.


pkg/controller/controller.go
```diff
@@ -274,17 +275,21 @@ func (m *CertificateApprover) reconcileCSR(csr certificatesv1.CertificateSigning
...
 
-	kubeletCA := m.getKubeletCA()
-	if kubeletCA == nil {
-		// This is not a fatal error.  The renewal authorization flow
-		// depending on the existing serving cert will be skipped.
-		klog.Errorf("failed to get kubelet CA")
-	}
-
-	if authorize, err := authorizeCSR(m.WorkloadClient, m.Config, machines, &csr, parsedCSR, kubeletCA); !authorize {
-		// Don't deny since it might be someone else's CSR
```
JoelSpeed, line 285: Lets make sure not to lose this context

pkg/controller/csr_check.go
```diff
@@ -64,6 +64,105 @@ var now = time.Now
...
+// nodeEgressIPs holds the egress IPs and CIDRs assigned to a node.
+type nodeEgressIPs struct {
+	IPs   []net.IP
+	CIDRs []*net.IPNet
+}
+
+// getNodeEgressIPs returns the egress IPs assigned to the given node.
+// It determines the network type and queries the appropriate CRD:
+// - OVNKubernetes: queries EgressIP objects from k8s.ovn.org/v1
+// - OpenShiftSDN: queries HostSubnet objects
```
JoelSpeed: SDN hasn't been supported since ~4.17 IIRC. We should [verify this](https://redhat-internal.slack.com/archives/CDCP2LA9L/p1783326016912359) and remove any SDN references now

pkg/controller/csr_check.go
```diff
@@ -182,66 +280,49 @@ func authorizeCSR(
...
-		if err := authorizeServingRenewal(nodeAsking, csr, servingCert, x509VerificationOpts); err != nil {
-			approvalErrors = append(approvalErrors, err)
-			klog.Infof("Could not use current serving cert for renewal: %v", err)
-			klog.Infof("Current SAN Values: %v, CSR SAN Values: %v",
-				certSANs(servingCert), csrSANs(csr))
-		} else {
-			// No error, the renewal is authorized.
-			return true, nil
+	// Use machine API authorization if the node has a machine.
+	machine, machineErr := machinehandlerpkg.FindMatchingMachineFromNodeRef(machines, nodeAsking)
```
JoelSpeed, line 266: Why are we reversing this order of machine vs renewing against existing SANs? Seems like an unrelated change we don't necessarily need to include here?

pkg/controller/csr_check.go
```diff
@@ -667,6 +687,14 @@ func subsetIPAddresses(cidrs []*net.IPNet, super, sub []net.IP) bool {
...
 }
 
+func makeIPSet(ips []net.IP) map[string]struct{} {
+	set := make(map[string]struct{}, len(ips))
+	for _, ip := range ips {
+		set[ip.String()] = struct{}{}
+	}
+	return set
+}
+
```
JoelSpeed, line 682: Use the [k8s.io/util/set](https://pkg.go.dev/k8s.io/utils/set) package rather than creating our own set semantics?

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20260317180604-743f664b82d1
 	github.com/openshift/controller-runtime-common v0.0.0-20260428152732-64ee174f5e2e
 	github.com/openshift/library-go v0.0.0-20260318142011-72bf34f474bc
+	github.com/ovn-kubernetes/ovn-kubernetes/go-controller v0.0.0-20260703130629-a6c3ed490571
```
JoelSpeed: This is an unfortunate dependency, is it necessary? Is it worth looking at unstructured instead?

mdbooth: Hah, that's exactly what the bot said. I said no. I think the ovn-kubernetes api package was fairly well isolated, no? IIRC this doesn't pull in anything other than apimachinery, which we've already got.

pkg/controller/csr_check.go
mdbooth: If you go back to the Jira that was my original plan. But @sdodson had asked for a 4.16 backport where it *is* required, and the code was already there so I left it in. I'm inclined to follow up with another patch removing SDN support back to 4.17.

pkg/controller/csr_check.go
mdbooth: We're not reversing it. We're *eliminating* it. There is no fallback any more. Either we have a Machine, in which case we do the Machine validation flow, or we don't, in which case we do the existing SAN validation flow.

This is relevant to EgressIPs because the IPI flow can reject a CSR with an EgressIP which was formerly present but is no longer present, but the existing SAN validation can't. If we allow a fallback back we remove that ability.

It's also interesting to note that if you look carefully at the tests, we haven't altered *any* existing test cases with this change, except by removing unrelated information from their error messages.

go.mod
mdbooth: I think Unstructured was viable here, I just prefer the relative safety of structured types whenever possible. I can look at this again, but IIRC it was just apimachinery, which I felt was safe enough.

pkg/controller/csr_check.go
```diff
@@ -297,19 +378,25 @@ func authorizeNodeClientCSR(c client.Client, machines []machinehandlerpkg.Machin
...
 	if !match {
-		return fmt.Errorf("CSR Subject Alternate Name values do not match current certificate")
+		return newCSRValidationError("CSR Subject Alternate Name values do not match current certificate")
+	}
+
+	// CSR IP addresses must be a subset of (current cert IPs + egress IPs).
+	allowedIPs := append(currentCert.IPAddresses, egress.IPs...)
+	if !subsetIPAddresses(egress.CIDRs, allowedIPs, csr.IPAddresses) {
+		return newCSRValidationError("CSR Subject Alternate Names includes unknown IP addresses")
 	}
```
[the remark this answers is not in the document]
mdbooth: How would a cert-chain/expiry failure be resolved by retrying?

go.mod
```diff
@@ -38,22 +39,22 @@ require (
...
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.61.0 // indirect
-	go.opentelemetry.io/otel v1.36.0 // indirect
+	go.opentelemetry.io/otel v1.41.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.34.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.34.0 // indirect
-	go.opentelemetry.io/otel/metric v1.36.0 // indirect
-	go.opentelemetry.io/otel/sdk v1.36.0 // indirect
-	go.opentelemetry.io/otel/trace v1.36.0 // indirect
+	go.opentelemetry.io/otel/metric v1.41.0 // indirect
+	go.opentelemetry.io/otel/sdk v1.39.0 // indirect
```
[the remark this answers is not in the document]
mdbooth: Unrelated.

go.mod
mdbooth: We discussed this. The resolution: on my recognizance 😅 I'm going to double check my assertion that the api-only import from ovn-kubernetes is as clean as I'm claiming, but if so then I'm ok with it.

pkg/controller/csr_check.go
mdbooth: We discussed this. @JoelSpeed additionally raised the question of whether this could impact platforms which use keepalived to manage a VIP, e.g. metal, vSphere, OpenStack.

To verify this, I checked the state of all e2e jobs running against the installer. I noted that, as expected, all UPI jobs implement an `approve_csrs` method to 'manually' approve CSRs. In contrast, only 1 IPI job currently does this: `openshift-e2e-vsphere-hybrid-env`. This job adds 2 BM nodes to a vSphere cluster. As the additional nodes will not be auto-approved, it approves the additional nodes.

We know that SAN-reapproval is not relevant to joining nodes, so this means that all IPI platforms under test are approved by CMA using the Machine-backed flow.

As for keepalived specifically, I checked what actually happens here. I created a vsphere cluster and found the control plane node which currently owned the external api VIP. I confirmed that the API VIP was added to br-ex as an additional IP address. I deleted kubelet's serving cert and restarted it, which caused it to request a new one. The new CSR:
* Did not include the api server VIP
* Was automatically approved

The reason was that the CCM had not added the VIP to the Node. The reason for this is that the CCM is only aware of IPs which have been added via infrastructure API calls. Keepalived is moving the VIP through L2 network shenanigans, and therefore isn't visible to the CCM.

By contrast, EgressIPs are added via infrastructure API calls, and are visible to the CCM. Therefore they must either be picked up by (ideally) the {Machine,Cluster}API provider, or CMA must be explicitly aware of them.

pkg/controller/csr_check.go
mdbooth: I gave this a thumbs up, but when I've gone to do it I've realised the problem: net.IP isn't `comparable` so you can't create a sets.Set for it. You have to do sets.New[string]() and then add strings to it manually (well, unless I also moves slices.Map here), so the locs here ends up being the same. Then you have to add related churn everywhere else an ipSet is expected to be map[string]struct{}. It ends up not being worth it 🤷

go.mod
mdbooth: Yep, this API is extremely clean: https://github.com/ovn-kubernetes/ovn-kubernetes/tree/master/go-controller/pkg/crd/egressip/v1

pkg/controller/csr_check.go
```diff
@@ -554,22 +585,30 @@ func getServingCert(c client.Client, nodeName string, ca *x509.CertPool) (*x509.
...
-		ServerName: host,
+	dialer := &tls.Dialer{
+		NetDialer: &net.Dialer{Timeout: 30 * time.Second},
+		Config: &tls.Config{
+			RootCAs:    ca,
+			ServerName: host,
+		},
 	}
 
 	klog.Infof("retrieving serving cert from %s (%s)", nodeName, kubelet)
```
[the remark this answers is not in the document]
mdbooth: This has been this way for at least 6 years. It's an unrelated change and I'm not convinced the security benefit is significant.

pkg/controller/csr_check.go
```diff
@@ -554,22 +585,30 @@ func getServingCert(c client.Client, nodeName string, ca *x509.CertPool) (*x509.
...
 
-	conn, err := tls.DialWithDialer(dialer, "tcp", kubelet, tlsConfig)
+	conn, err := dialer.DialContext(ctx, "tcp", kubelet)
 	if err != nil {
 		return nil, err
 	}
 
 	defer conn.Close()
```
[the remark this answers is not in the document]
mdbooth: Unrelated change.

pkg/controller/csr_check.go
JoelSpeed: That comment is a year old, do we still intend to backport this all the way to 4.16? I would expect us to only backport this as far as 4.18 which is the oldest, currently in maintenance release.

Sustaining may wish to backport to 4.16 to support the EUS term 2 customers that are using 4.16 there, perhaps we should get their impression on whether this woud be something they want to backport?

pkg/controller/csr_check.go
```diff
@@ -64,6 +64,105 @@ var now = time.Now
...
+	if err := c.Get(ctx, client.ObjectKey{Name: networkClusterName}, network); err != nil {
+		return result, fmt.Errorf("could not fetch cluster network: %v", err)
+	}
+
+	switch network.Status.NetworkType {
+	case string(operatorv1.NetworkTypeOVNKubernetes):
+		egressIPList := &egressipv1.EgressIPList{}
+		if err := c.List(ctx, egressIPList); err != nil {
+			return result, fmt.Errorf("could not list EgressIP objects: %v", err)
+		}
```
RadekManak, line 99: Do you have RBAC for egressIP?

pkg/controller/csr_check.go
```diff
@@ -430,13 +458,15 @@ func authorizeServingCertWithMachine(machines []machinehandlerpkg.Machine, req *
...
+				continue
+			}
 
 			//TODO: set annotation/emit event here.
-			// return error so we requeue, in case machine network is out of date
-			// for some reason
-			klog.Errorf("%v: IP address '%s' not in machine addresses: %s", req.Name, san, strings.Join(attemptedAddresses, " "))
-			return fmt.Errorf("IP address '%s' not in machine addresses: %s", san, strings.Join(attemptedAddresses, " "))
+			klog.Errorf("%v: IP address '%s' not in machine addresses or egress IPs: %s", req.Name, san, strings.Join(attemptedAddresses, " "))
+			return newCSRValidationError("IP address '%s' not in machine addresses or egress IPs: %s", san, strings.Join(attemptedAddresses, " "))
```
RadekManak: Are the IPs and DNS already available when the CSR is first reconciled? CSRValidationError is being handled as permanent failure. Previously this returned error and requeued.

mdbooth: This is ok because:
* For the initial serving cert there are no egress IPs and we know Machine addresses are up to date
* For rotations, kubelet will retry after 15 minutes by creating a new CSR

pkg/controller/csr_check.go
mdbooth: Nope 😅

PR #312, opened 2026-07-28
OCPBUGS-100160: Fix status controller unit-test teardown race

RadekManak:
### Cause
With `--randomize-all`, a status-controller case that leaves `RELEASE_VERSION` unset can return from `Run` while a queue worker is still inside `statusAvailable`. That lingering sync recreates/updates the ClusterOperator as `unknown` after the next case has already started. The suite's `Eventually` only required a non-empty `Versions` slice, so it accepted the stale value.

### Consequence
Unrelated PRs flake on `ci/prow/unit` in ~9ms with `Expected unknown to equal a_cvo_given_version`:
- https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-machine-approver/310/pull-ci-openshift-cluster-machine-approver-main-unit/2082042738303307776
- https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-machine-approver/304/pull-ci-openshift-cluster-machine-approver-main-unit/2059912951690694656

### Fix
- On stop: `queue.ShutDown()` then `WaitGroup.Wait()` so workers finish before `Run` returns; also `ShutDown` on the cache-sync early-return path.
- Prefer `versionGetter` for `Versions` (fallback to `RELEASE_VERSION`) so status matches the Available / `SetVersion` path.
- `Eventually` waits until `Versions[0].Version == expectedVersion`.

### Result
Cross-case CO writes after teardown are blocked, and a stale non-empty `Versions` slice can no longer false-pass the suite.

## Context & Patterns

**Why WaitGroup after ShutDown, not just defer ShutDown** — Closing `stopCh` used to shut down the queue and return from `Run` while a worker could still be inside `statusAvailable`. In the suite, `AfterEach` treats `Run` return as "idle" and deletes the ClusterOperator; a late sync then recreates it as `unknown` for the next case. Shutting down the queue first, then waiting on workers, makes "controller stopped" mean no further CO writes.

**Why Eventually asserts the exact version** — With `--randomize-all`, a prior case that left `RELEASE_VERSION` unset can leave a stale non-empty `Versions` slice. Waiting only for `len(Versions) > 0` accepts that stale `unknown` in a few milliseconds. Waiting for `Versions[0].Version == expectedVersion` turns cross-case contamination into a timeout instead of a false pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the cluster operator status controller’s shutdown behavior to ensure background workers complete cleanly.
  * Enhanced cluster operator status version reporting to prefer the configured operator version when available, with a fallback to the release version.
  * Made operator availability/status messaging consistently reflect the computed operator version.
* **Tests**
  * Updated status tests to repeatedly verify the reported cluster operator version matches the expected value before asserting success.

### DIFF
--- a/status.go
+++ b/status.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	osconfigv1 "github.com/openshift/api/config/v1"
@@ -95,8 +96,15 @@ func (c *statusController) runWorker() {
 func (c *statusController) Run(threadiness int, stopCh chan struct{}) {
 	defer utilruntime.HandleCrash()
 
-	// Let the workers stop when we are done
-	defer c.queue.ShutDown()
+	// Shut down the queue then wait for workers on every exit path (cache-sync
+	// failure or stopCh) so an in-flight sync cannot recreate ClusterOperator
+	// state after Run returns.
+	var wg sync.WaitGroup
+	defer func() {
+		c.queue.ShutDown()
+		wg.Wait()
+	}()
+
 	klog.Info("Starting cluster operator status controller")
 
 	go c.clusterOperatorInformer.Run(stopCh)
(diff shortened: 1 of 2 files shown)


damdo (commented):
/approve
/lgtm

